PR #5, opened 2017-07-06
Improved Doxygen setup

dominikwerder:
Improves how we use Doxygen to build the documentation, specifically to avoid issues in Doxygen when linking to other md files.


dominikwerder:
Sorry, wrong branch.

PR #11, opened 2017-07-07
Ansible install

michele-brambilla:
**nexus-to-kafka** can be installed using the ansible playbook. Required libraries are installed if not present (paths are defined in the role variables).
The installation (on our testing machines) requires _libz_ to be added to src/CMakeLists.txt "libraries_common"

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,12 @@ environment variables here:
 
 ### Using Ansible
 
+Install using the playbook:
+
+```
+ansible-playbook -i hosts kafka-to-nexus.yml
+```
+
 The filewriter can be installed using the ansible playbook defined in
 `ansible`. The file `roles/kafka-to-nexus/defaults/main.yml` defines
 the variables used during installation. The variables `<dep>_src` and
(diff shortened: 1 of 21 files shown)


matthew-d-jones:
Hi Michele, it looks good, I'd like to try it. Could you add a couple of lines to the documentation to describe what I should edit to specify a target and how to run the playbook?

matthew-d-jones:
I had a problem getting this to run, turns out the make module requires Ansible 2.1, it might be nice to add the following to `kafka-to-nexus.yml` to check the Ansible version.
```
pre_tasks:
      - name: Verify we have required version of Ansible.
        assert:
          that: "ansible_version.full | version_compare('2.1', '>=')"
          msg: >
            "You must update Ansible to at least 2.1 to use this playbook."
```

I also suggest adding a comment to the readme to run with `ansible-playbook -i hosts kafka-to-nexus.yml` for idiots like me who always forget to specify the inventory file and are then confused when it complains there are no target hosts.

Is the target machine assumed to be running a standard ESS image? I had several missing dependencies on a clean CentOS 7 VM: git, cmake, gcc-c++.

michele-brambilla:
I added the `pre_task` suggestion. After a couple of minor changes it compiles on the ESS image.

PR #12, opened 2017-07-10
Use hash map for the set of Source instances

dominikwerder:
Valgrind/Callgrind showed hot spot if number of streamed sources gets large.
Using a hash map scales better.

### DIFF
--- a/src/DemuxTopic.h
+++ b/src/DemuxTopic.h
@@ -5,6 +5,7 @@
 #include "json.h"
 #include <functional>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace BrightnESS {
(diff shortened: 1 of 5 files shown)


src/DemuxTopic.h
```diff
@@ -30,10 +31,12 @@ class DemuxTopic : public TimeDifferenceFromMessage, public MessageProcessor {
   ProcessMessageResult process_message(char *msg_data, int msg_size);
   /// Implements TimeDifferenceFromMessage.
   DT time_difference_from_message(char *msg_data, int msg_size);
-  std::vector<Source> &sources();
-  template <typename... Args> Source &add_source(Args &&... args) {
-    _sources.emplace_back(std::forward<Args>(args)...);
-    return _sources.back();
+  std::unordered_map<std::string, Source> &sources();
+  Source &add_source(Source &&source) {
+    using std::move;
```
matthew-d-jones: This is a very minor point but I think using `using` here just adds a line of code for no readability improvement.

PR #14, opened 2017-07-12
Fix segfault due to return of reference to stack variable

dominikwerder:


dominikwerder:
The reporting of the status seems not yet functional.  I talked to Michele and he will refactor the reporting of the status of the StreamMaster class, which would also imply again new unit tests. So I think we could just merge this fix for now?

PR #16, opened 2017-07-24
Parse commands given in configuration file

dominikwerder:
Add the possibility to issue file writer commands from the configuration file. Those commands are processed before any command is taken from the Kafka command topic.

### DIFF
--- a/src/MainOpt.cxx
+++ b/src/MainOpt.cxx
@@ -11,6 +11,7 @@ using uri::URI;
 
 int MainOpt::parse_config_file(std::string fname) {
   using namespace rapidjson;
+  using std::move;
   if (fname == "") {
     LOG(3, "given config filename is empty");
     return -1;
@@ -42,6 +43,13 @@ int MainOpt::parse_config_file(std::string fname) {
         master_config.kafka.emplace_back(m.name.GetString(),
                                          fmt::format("{}", m.value.GetInt()));
       }
+    }
+  }
+  if (auto a = get_array(d, "commands")) {
+    for (auto &e : a.v->GetArray()) {
+      Document js_command;
+      js_command.CopyFrom(e, js_command.GetAllocator());
+      master_config.commands_from_config_file.push_back(move(js_command));
     }
   }
   return 0;
--- a/src/MainOpt.cxx
+++ b/src/MainOpt.cxx
@@ -12,7 +12,7 @@ using uri::URI;
 int MainOpt::parse_config_file(std::string fname) {
   using namespace rapidjson;
   using std::move;
-  if (fname == "") {
+  if (fname.empty()) {
     LOG(3, "given config filename is empty");
     return -1;
   }
--- a/src/MainOpt.cxx
+++ b/src/MainOpt.cxx
@@ -11,7 +11,6 @@ using uri::URI;
 
 int MainOpt::parse_config_file(std::string fname) {
   using namespace rapidjson;
-  using std::move;
   if (fname.empty()) {
     LOG(3, "given config filename is empty");
     return -1;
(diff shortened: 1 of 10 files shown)


src/MainOpt.cxx
```diff
@@ -11,6 +11,7 @@ using uri::URI;
 
 int MainOpt::parse_config_file(std::string fname) {
   using namespace rapidjson;
+  using std::move;
```
matthew-d-jones: Don't use `using` for something you only use once.

src/MainOpt.cxx
```diff
@@ -11,6 +11,7 @@ using uri::URI;
...
   if (fname == "") {
```
matthew-d-jones: The `empty` method should be used to check for emptiness.

src/Master.cxx
```diff
@@ -23,7 +23,15 @@ void Master::handle_command_message(std::unique_ptr<KafkaW::Msg> &&msg) {
...
+  CommandHandler command_handler(this, config.config_file);
+  command_handler.handle(cmd);
+}
+
 void Master::run() {
+  for (auto &cmd : config.commands_from_config_file) {
+    this->handle_command(cmd);
+  }
   command_listener.start();
   if (_cb_on_connected)
```
matthew-d-jones: This does an implicit cast to bool, instead compare with null: `if (_cb_on_connected != nullptr)`

src/Master.cxx
```diff
@@ -23,7 +23,15 @@ void Master::handle_command_message(std::unique_ptr<KafkaW::Msg> &&msg) {
   command_handler.handle({(char *)msg->data(), (int32_t)msg->size()});
 }
 
+void Master::handle_command(rapidjson::Document &cmd) {
+  CommandHandler command_handler(this, config.config_file);
+  command_handler.handle(cmd);
+}
+
 void Master::run() {
+  for (auto &cmd : config.commands_from_config_file) {
```
matthew-d-jones: Could this be `auto const &cmd`? It would require a few method signatures to change, including `handle_command`.

src/helper.h
```diff
@@ -31,6 +31,12 @@ struct get_json_ret_int {
   int v;
 };
 
+struct get_json_ret_array {
+  operator bool() const;
```
matthew-d-jones: This and other operators in this file should be marked `explicit` to avoid unintentional implicit conversions.

src/helper.h
```diff
@@ -39,6 +45,7 @@ struct get_json_ret_object {
 
 get_json_ret_string get_string(rapidjson::Value const *v, std::string path);
 get_json_ret_int get_int(rapidjson::Value const *v, std::string path);
+get_json_ret_array get_array(rapidjson::Value const &v, std::string path);
 get_json_ret_object get_object(rapidjson::Value const &v, std::string path);
 std::string pretty_print(rapidjson::Document const *v);
```
matthew-d-jones: I don't think `pretty_print` is ever used, it should be removed or a comment added to explain why it is here.

src/helper.h
matthew-d-jones: On the subject of being more explicit, I don't like the name `helper` filename. If the filename is generic because there are helper functions for different things in here then consider splitting them into multiple files. Alternatively, if they are not needed by multiple classes they could go in the file and namespace of the class they help.

dominikwerder:
The reason for this change is that it adds support for writing only a single
file, followed by process termination.  The stop condition (time, written size
or something else) will come after the improved status reporting of the
underlying Kafka stream handling has landed in a different PR.
This change also allows for some simplified testing which I find useful at
least for development.

src/helper.h
dominikwerder: Right, I'll split the helper file in a later PR.

PR #20, opened 2017-07-27
Fix remaining compiler warnings

dominikwerder:
Tested on GCC 4.9.3 and clang master (the to-be 5.0).


dominikwerder:
:+1:

dominikwerder:
It does pass though with `-Wno-variadic-macros -Wno-gnu-zero-variadic-macro-arguments -Wno-gnu-include-next`

PR #21, opened 2017-07-31
Added StreamMaster status information

### LINKED ISSUE #26: StreamMaster constructor does not return if Kafka broker unreachable
The constructor of class `StreamMaster` does not return if the Kafka broker is unavailable.

If the class `Master` creates a `StreamMaster` without available broker, then control is not returned to the `Master` class, the program hangs and `Master` can not even detect that because its thread got 'stolen' :smiley:

### LINKED ISSUE #28: search_backward mentioned in readme
`search_backward` is still mentioned in the README despite having been removed from the codebase.

michele-brambilla:
StreamMaster exposes a queue which contains informations on its status and on each streamer:

- StreamMaster stopped, running or "has finished" 
- for each Streamer number of messages, errors and byte received
- for each Streamer average message size and frequency (plus rms)

The StreamMaster status can be converted in a rapidJSON Document (or printed on the screen for debugging)

Closes #26
Closes #28


matthew-d-jones:
If I understand correctly, `StreamStatus` is a queryable log of the status and some throughput statistics. What are the reasons for implementing this functionality in the file writer, rather than publishing status messages to a topic to be monitored by an external service?

michele-brambilla:
For the moment we wanted the Master to be aware of these information. I agree with you that it makes sense to have them in Kafka as well, not sure if the StreamMaster should public or the Master (e.g. do we want to add other information from it?) should.
Furthermore the Master would know informations about the StreamMaster (mainly the stopped/running/finished status), so for the moment we assume it to be in charge of reading all these informations.

matthew-d-jones:
Why does `Master` need this information? Does it have something to do with load balancing?

dominikwerder:
Currently, the `Master` class has no way to tell whether the `StreamMaster` either is still running, or has finished its job because it reached a preset time limit, or has encountered an error. Therefore `Master` currently does not know when to stop and deallocate a `StreamMaster`. Therefore, we talked about that `StreamMaster` should expose a method where `Master` can poll the current status, even though I couldn't find that method when I just skimmed the PR?

Regarding publishing the status: We discussed that the `StreamMaster` instances could expose the status updates via some ringbuffer/queue/... such that the `Master` class can consolidate them and send off to Kafka such that less Kafka connections are needed.

matthew-d-jones:
The `librdkafka` API is thread-safe, if the only reason that `Master` needs this information is to send messages from a Kafka producer, then it would be far more simple to have joint ownership of the producer and send messages from `StreamMaster`.

michele-brambilla:
Proposal: we get rid of the queue and associate a producer to each ``StreamMaster``. This will make easier (if we would) go multiprocess, while the number of spawned (kafka) threads will be under control, assuming one (or few) ``StreamMaster`` per instrument/beamline

matthew-d-jones:
@michele-brambilla Is there a reason why one producer for all `StreamMaster`s doesn't work?

michele-brambilla:
In this case we should rely on ``Master`` in some way (and if I understand we are trying to avoid that). Either it is the owner of the producer or we need one ``StreamMaster`` to be and share it among other ``StreamMaster``. The latter can be done, of course, but still relies on the ``Master`` to forward the producer to the ``Streamer``s.
Furthermore, when we will go multiprocess I guess the minimum requirement will be "one producer per process". I don't think share a producer among processes would be a great idea..

matthew-d-jones:
@michele-brambilla As I said previously, the `librdkafka` API is thread-safe. There should be no problem with having one producer shared between `StreamMaster`s. `Master` can create a producer and share ownership (`std::shared_ptr`) with the `StreamMaster`s.

> In this case we should rely on Master in some way (and if I understand we are trying to avoid that).

My objection to feeding messages back to `Master` through a queue or ringbuffer is not about avoiding relying on `Master`, but that it seems to be unnecessarily complex and has the memory overhead of the buffers.

michele-brambilla:
Let me make clear that I'm not against one single producer belonging to ``Master``, it makes absolutely sense. Nevertheless i fear that the point with multiprocess (not multithreading) still exists..
Let's put this way, if it is OK with Dominik we can start with the ``Master`` to be in charge of producing these messages; if we later realise that would be better each ``StreamMaster`` to be in charge it should be straightforward to modify.

matthew-d-jones:
> Nevertheless i fear that the point with multiprocess (not multithreading) still exists..

I don't understand this concern. Could you explain what problem multithreading introduces?

michele-brambilla:
Multithreading does not introduce any problem, it's trivial. My concern is about spawning different processes; if at some point we figure out that a single machine cannot support the whole writing burden of writing to disk we have to go that way.

matthew-d-jones:
Ah, would you have only one `Master` on one machine and multiple `StreamMaster`s on different machines in that circumstance?

michele-brambilla:
Let's say that at least we are considering the possibility, and at the moment we'd like to keep everything as flexible as possible. But as I said, it shouldn't be difficult to migrate the producer from ``Master`` to ``StreamMaster`` (or the other way around). Just keep in mind that it will be affected by the approach to multiprocess parallelisation.

matthew-d-jones:
Fair enough, I understand the concern now. I would suggest adding complexity only when you are sure you need it (the YAGNI approach). I would create a single producer in `Master` and for `StreamMaster`s to hold a `shared_ptr` for now. Then if `StreamMaster`s are in separate processes in the future they can simply create their own producer in their constructor instead of accepting the pointer.
If you instead prefer to already have a separate producer in each `StreamMaster` then I won't strongly object.

Do you want me to go ahead and look at the code in this PR, or will you make changes based on the discussion we've had?

michele-brambilla:
Let me make changes before. BTW I'm also adding a GELF output format so that we can eventually send these information to Graylog via Kafka (what do you think about it?). I will send you a message when everything is ready.

matthew-d-jones:
I'm not very keen on GELF because it makes it more difficult for any consumer other than Graylog to interpret the message. If you want to send a message to Graylog then you already have Jonas's library for that.
I think we should be considering sending all log messages through Kafka, and possibly using Grafana instead of Graylog as Dominik, myself and, judging by https://confluence.esss.lu.se/display/DMSC/Event+Formation, Morten are keen on that. Maybe we can discuss this at the meeting tomorrow?

michele-brambilla:
The idea is to send the message via Kafka in any case. Graylog provides a "Raw/Plaintext Kafka" input which we already tested some time ago. After the last messages we had I was considering to use using JSON messages + Graylog extractors (if we want to go for Graylog). Or simply use Grafana.

michele-brambilla:
Implementation changes after discussions:

- `Master` shares a Kafka producer with each the `StreamMaster` to be used for streaming status informations
- each `Streamer` collects informations on the status, the `StreamMaster` gather all these and translate into a JSON message
- the `StreamMaster` exposes a `has_finished` message to inform the `Master` that all the `Streamer`s are done writing or that any error occurred

Further changes:

- each `Streamer` connection is now asynchronous. No matter ho long it takes to connect, after instantiation the flow controls returns immediately to the `Master`
- a failure in the initial connection (_e.g._ typo in the broker address) is considered critical. RdKafka is then in charge of handle any successive failure
- RdKafka::Conf and Streamer option are now distinct constructor parameters
- updated README and removed dummy target

src/StreamMaster.hpp
```diff
@@ -84,90 +104,177 @@ template <typename Streamer, typename Demux> struct StreamMaster {
...
+      auto value = Status::pprint<Status::JSONStreamWriter>(status);
+      if (!report_producer_) {
+        LOG(1, "ProucerTopic error: can't produce StreamMaster status report")
+        runstatus = Error::statistics_failure;
+        return;
+      }
+      report_producer_->produce(reinterpret_cast<unsigned char *>(&value[0]),
+                                value.size());
+    }
+    return;
```
matthew-d-jones: `return` is redundant

src/StreamMaster.hpp
```diff
@@ -84,90 +104,177 @@ template <typename Streamer, typename Demux> struct StreamMaster {
...
+  }
+
+  void stop_impl(void) {
```
matthew-d-jones: `void` is redundant

src/StreamMaster.hpp
```diff
@@ -84,90 +104,177 @@ template <typename Streamer, typename Demux> struct StreamMaster {
...
+      if (!report_producer_) {
+        LOG(1, "ProucerTopic error: can't produce StreamMaster status report")
+        runstatus = Error::statistics_failure;
+        return;
+      }
+      report_producer_->produce(reinterpret_cast<unsigned char *>(&value[0]),
+                                value.size());
+      std::this_thread::sleep_for(std::chrono::milliseconds(delay));
+    }
+    // temination message
```
matthew-d-jones: "temination" -> "termination" typo

src/StreamMaster.hpp
```diff
@@ -9,10 +9,15 @@
...
+#include <queue>
 #include <thread>
 
 #include "DemuxTopic.h"
 #include "FileWriterTask.h"
+#include "KafkaW.h"
+#include "Status.hpp"
+#include "StatusWriter.hpp"
 #include "logger.h"
 #include "utils.h"
```
matthew-d-jones: I realise that it is not in your changes, but on line 24 there is a declaration (`struct Streamer;`) which should be removed. The `Streamer` being used is in the `FileWriter` namespace so this one must be a mistake.

src/tests/status_test.cxx
```diff
@@ -0,0 +1,117 @@
...
+      continue;
+    }
+    double n = normal(generator);
+    s.add_message(n);
+    x += n;
+  }
+  return;
+}
+
+void continous_collect(Status::StreamerStatus &s, bool &terminate) {
```
matthew-d-jones: "continous" -> "continuous" throughout this file

src/tests/status_test.cxx
```diff
@@ -0,0 +1,117 @@
+#include <Status.hpp>
+
+#include <gtest/gtest.h>
+
+#include <random>
```
matthew-d-jones: `random` isn't used

src/tests/status_test.cxx
```diff
@@ -0,0 +1,117 @@
...
+  for (int i = 0; i < n_messages; ++i) {
+    if (i % error_ratio == 0) {
+      s.error();
+      continue;
+    }
+    double n = normal(generator);
+    s.add_message(n);
+    x += n;
+  }
+  return;
```
matthew-d-jones: `return` is redundant

src/tests/status_test.cxx
```diff
@@ -0,0 +1,117 @@
...
+  double x{0};
+  std::thread update(collect, std::ref(s_), std::ref(x));
+  update.join();
+  std::thread test(fetch, std::ref(s_), std::ref(x));
+  test.join();
+}
+
+TEST(StreamerStatus, ContinousCounters) {
+  Status::StreamerStatus s_;
+  bool stop = false;
```
matthew-d-jones: You're relying on `bool` being atomic, which is _probably_ the case but is not guaranteed by the standard. Better to make this a `std::atomic<bool>`.

src/tests/status_test.cxx
```diff
@@ -0,0 +1,117 @@
...
+TEST(StreamerStatus, Counters) {
+  Status::StreamerStatus s_;
+  double x{0};
+  std::thread update(collect, std::ref(s_), std::ref(x));
+  update.join();
+  std::thread test(fetch, std::ref(s_), std::ref(x));
+  test.join();
+}
+
+TEST(StreamerStatus, ContinousCounters) {
```
matthew-d-jones: One of the most important roles of tests is to document what the programmer intends to happen. The tests provide confidence that the what the programmer intends to happen actually does. To catch the other possibility for bugs, that there is a mistake in the intention, it is important to make it obvious to someone reading the tests what is being tested. For this reason I like more explicit test names and use of snake_case for readability, for example "test_status_counters_increase_over_time". The BDD style tests that Mark is keen on are also a great way to achieve this.

src/tests/streammaster_test.cxx
```diff
@@ -141,6 +141,37 @@ TEST(Streammaster, StartTime) {
   sm.start_time(ESSTimeStamp(10));
 }
 
+// TEST(Streammaster, Statistics) {
```
matthew-d-jones: I guess the commented out implementation and test are here because they are part of functionality that you intend to include in a future PR? If so, I think it would be better to remove the commented code from here, make a branch from this one and add it, uncommented, to the new branch.

src/utils.h
```diff
@@ -34,21 +34,34 @@ template <typename T, typename PHANTOM> struct StrongType {
...
 
-enum StatusCode {
-  NO_ERROR = 1000,
-  RUNNING = 1,
-  STOPPED = 0,
-  ERROR = -1,
-};
+
+
+/* enum StatusCode { */
```
matthew-d-jones: Can this be removed rather than commented out?

src/utils.h
```diff
@@ -34,21 +34,34 @@ template <typename T, typename PHANTOM> struct StrongType {
 struct OffsetType {};
 struct PartitionType {};
 struct Timestamp {};
-struct FileWriterErrorCode {};
+struct FileWriterErrorType {};
+ 
+ struct StreamerErrorType{};
```
matthew-d-jones: Extra spaces at the start of some lines, is that intentional?

README.md
```diff
@@ -321,24 +314,36 @@ Both receive the message payload and size. Return values are ProcessMessageResul
...
+converts the error code into a human readable string). Possible values:
+
+| Code                 |
+|----------------------|
+|  not_started         | 
+|  running             |
+|  has\_finished       |
+|  streamer\_error     |
+|  statistics\_failure |
+|  streammaster\_error |
```
dominikwerder: Would it make sense to put internal technical documentation into (doxygen-enabled) comments into the source files, and leave README for information relevant for the users of the application?

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include <cmath>
+#include <iostream>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+#include "utils.h"
+
+namespace FileWriter {
+class Streamer;
```
dominikwerder: Streamer is originally defined as `struct`, forward declaring as class causes a warning.

src/StatusWriter.hpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
+// #include "schemas/fws0_fwr_status_generated.h"
+#include "rapidjson/document.h"
+
+namespace FileWriter {
+namespace Status {
+
+class StreamMasterStatus;
+class StreamerStatusType;
+class StreamerStatisticsType;
```
dominikwerder: `StreamerStatisticsType` is struct originally, causes warning.

src/StreamMaster.hpp
```diff
@@ -84,90 +104,177 @@ template <typename Streamer, typename Demux> struct StreamMaster {
...
 
   FileWriterTask const &file_writer_task() const { return *_file_writer_task; }
 
+  const StreamMasterError &status() {
+    for (auto &s : streamer) {
+      if (s.second.runstatus().value() < 0) {
+        runstatus = Error::streamer_error;
+      }
+    }
+    return StreamMasterError{runstatus.load()};
```
dominikwerder: Return ref to temporary, looks like UB to me. Could a simple return by value be an alternative?

src/StreamMaster.hpp
```diff
@@ -84,90 +104,177 @@ template <typename Streamer, typename Demux> struct StreamMaster {
...
   }
-  ErrorCode stop_streamer(Streamer &s) { return s.closeStream(); }
+  typename Streamer::Error stop_streamer(Streamer &s) {
+    return s.closeStream();
+  }
 
   void run() {
     using namespace std::chrono;
+    runstatus = Error::running;
     system_clock::time_point tp, tp_global(system_clock::now());
```
dominikwerder: Is `tp_global` still used?

src/StreamMaster.hpp
```diff
@@ -21,31 +26,39 @@ struct FileWriterCommand;
...
       streamer[d.topic()].n_sources() = d.sources().size();
     }
   };
 
-  StreamMaster(
-      std::string &broker, std::unique_ptr<FileWriterTask> file_writer_task,
-      std::vector<std::pair<std::string, std::string>> kafka_options = {},
-      const RdKafkaOffset &offset = RdKafkaOffsetEnd)
-      : demux(file_writer_task->demuxers()), do_write(false), _stop(false),
+  StreamMaster(std::string &broker,
```
dominikwerder: Minor thing, but if taken by `&`, it could be `const` here as well.

src/Streamer.cxx
```diff
@@ -84,225 +80,240 @@ int FileWriter::Streamer::get_topic_partitions(const std::string &topic) {
...
+  kafka_options.push_back(
+      FileWriter::Streamer::option_t{"api.version.request", "true"});
+  kafka_options.push_back(
+      FileWriter::Streamer::option_t{"group.id", topic_name});
+
+  connect_ = std::thread([&] {
+    this->connect(topic_name, kafka_options, filewriter_options);
+    return;
+  });
+  std::this_thread::sleep_for(milliseconds(10));
```
dominikwerder: Why a `sleep` in the constructor?

michele-brambilla:
Let me check with :white_check_mark: the items completed

README.md
```diff
@@ -301,16 +301,9 @@ According to the design the Streamer connects to Kafka (other
...
-  informations. ```search_backward(f)``` implements some algorithm that uses the
-  function ```f``` to find the older message with good timestamp. Different
-  sources can have data in different point of the queue: Source has to discard
-  invalid (according to timestamp) data
-* if the broker is not valid (_e.g._ a change of IP address) it should notify
-  the FileMaster, retrieve the new configuration and reconnect
-
+* initial timestamp is specified using ``set_start_time``
+* connection to the Kafka broker is nonblocking. If the broker address is invalid returns an error
+* Kafka::Config ans streamer options can be passed using the constructor ``kafka_options`` and ``filewriter_options`` respectively.
```
michele-brambilla: :white_check_mark:

README.md
```diff
@@ -321,24 +314,36 @@ Both receive the message payload and size. Return values are ProcessMessageResul
...
-The StreamMaster receives the array of DemuxTopic from FileWriterCommand and
-instantiates the Streamer array according to the topics. Eventually retrieves
-the list of brokers from Kafka.
+## StreamMaster 
+
+The StreamMaster receives the array of DemuxTopic from
+FileWriterCommand and instantiates the Streamer array according to the
+topics. Eventually retrieves the list of brokers from Kafka.
+
+* `start_time` and `stop_time` can be used to set a the timestamp of
```
michele-brambilla: :white_check_mark:

README.md
```diff
@@ -321,24 +314,36 @@ Both receive the message payload and size. Return values are ProcessMessageResul
...
+the fist and last event to be written (caution: actually events
```
michele-brambilla: :white_check_mark:

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+public:
+  StreamerStatusType();
+  StreamerStatusType(const StreamerStatusType &other);
+  StreamerStatusType(StreamerStatusType &&other) noexcept = default;
+
+  ~StreamerStatusType() = default;
+
+  StreamerStatusType &operator=(const StreamerStatusType &other);
+  StreamerStatusType &operator=(StreamerStatusType &&other);
```
michele-brambilla: :white_check_mark:

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+  double errors{0.0};
+  double bytes2{0.0};
+  double messages2{0.0};
+};
+
+// Data type for extracting statistics about the streamer
+struct StreamerStatisticsType {
+  double size_avg{0}, size_std{0};
+  double freq_avg{0}, freq_std{0};
+  int partitions{0};
```
michele-brambilla: The number of partition  per topic was an initial request from Dominik. At the moment is not implemented so I will remove the variable, eventually to be restored later.
 :white_check_mark:

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+};
+
+class StreamMasterStatus {
+  friend class StdIOWriter;
+  friend class JSONWriter;
+  friend class FlatbuffersWriter;
+
+public:
+  StreamMasterStatus() = default;
+  StreamMasterStatus(const int &value) : status(value){};
```
michele-brambilla: :white_check_mark:

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+  StreamMasterStatus &push(const std::string topic_name,
```
michele-brambilla: :white_check_mark:

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+  StreamerStatus(const StreamerStatus &other)
+      : current{other.current}, last{other.last}, last_time{other.last_time},
+        partitions{other.partitions}, run_status_(other.run_status_) {}
+
+  void add_message(const double &bytes);
+  void error() { current.errors++; }
+  void error(const StreamerError &value) {
+    run_status_ = value;
+    error();
+    return;
```
michele-brambilla: I had this thing at some moment to mark functions taking no argument with foo(void) and end functions returning void with return. My bad, I will get rid of all of them.

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
+#include <map>
+
+#include "Status.hpp"
+
+#include <rapidjson/document.h>
+#include <rapidjson/stringbuffer.h>
+#include <rapidjson/writer.h>
```
michele-brambilla: :white_check_mark:

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
...
+
+FileWriter::Status::StreamerStatusType::StreamerStatusType() {}
```
michele-brambilla: :white_check_mark:

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
...
+  return *this;
+}
+FileWriter::Status::StreamerStatusType FileWriter::Status::StreamerStatusType::
+operator+(const FileWriter::Status::StreamerStatusType &other) {
+  FileWriter::Status::StreamerStatusType result(*this);
+  return std::move(result += other);
+}
+void FileWriter::Status::StreamerStatusType::reset() {
+  bytes = messages = errors = bytes2 = messages2 = 0;
+  return;
```
michele-brambilla: (see before)

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
...
+  } else {
+    st.size_avg = -1.;
+    st.size_std = -1.;
+    st.freq_avg = -1.;
+    st.freq_std = -1.;
+  }
+  last += current;
+  last_time = t;
+  current.reset();
+  return std::move(st);
```
michele-brambilla: :white_check_mark:

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
...
+
+FileWriter::Status::StreamerStatusType::StreamerStatusType() {}
+
+FileWriter::Status::StreamerStatusType::StreamerStatusType(
+    const FileWriter::Status::StreamerStatusType &other)
+    : bytes{other.bytes}, messages{other.messages}, errors{other.errors},
+      bytes2{other.bytes2}, messages2{other.messages2} {}
+
+FileWriter::Status::StreamerStatusType &FileWriter::Status::StreamerStatusType::
+operator=(const FileWriter::Status::StreamerStatusType &other) {
```
michele-brambilla: :white_check_mark:

src/StreamMaster.hpp
michele-brambilla: :white_check_mark:

src/StatusWriter.cxx
```diff
@@ -0,0 +1,144 @@
...
+std::string JSONStreamWriter::write(const StreamMasterStatus &data) {
+  auto value = write_impl(data);
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> w(buffer);
+  value.Accept(w);
+  std::string s{buffer.GetString()};
+  return std::move(s);
+}
+
+// flatbuffers::Offset<StatusInfo>
```
michele-brambilla: :white_check_mark: (in the header file too)

src/tests/status_test.cxx
michele-brambilla: :white_check_mark:

src/StreamMaster.hpp
michele-brambilla: :white_check_mark:

matthew-d-jones:
@michele-brambilla 
This PR will fix issues #26 and #28, if you add the following lines to the PR description at the top of this page then github will automatically close the issues when this PR is merged:
Closes #26 
Closes #28

src/Status.hpp
```diff
@@ -0,0 +1,155 @@
...
+  StreamerStatusType &operator+=(const StreamerStatusType &other);
+  StreamerStatusType &operator-=(const StreamerStatusType &other);
+  StreamerStatusType operator+(const StreamerStatusType &other);
+
+  void reset();
+
+  double bytes{0.0};
+  double messages{0.0};
+  double errors{0.0};
+  double bytes2{0.0};
```
michele-brambilla: Added more explanatory names and some comments. Same for ``StreamerStatisticsType``

src/tests/status_test.cxx
michele-brambilla: :white_check_mark:

src/utils.h
michele-brambilla: :white_check_mark:

src/Status.cxx
```diff
@@ -0,0 +1,149 @@
...
+static const std::map<const int, std::string> stream_master_error_lookup_{
+    {FileWriter::Status::StreamMasterErrorCode::not_started, "not_started"},
+    {FileWriter::Status::StreamMasterErrorCode::running, "running"},
+    {FileWriter::Status::StreamMasterErrorCode::has_finished, "has_finished"},
+    {FileWriter::Status::StreamMasterErrorCode::statistics_failure,
+     "statistics_failure"},
+    {FileWriter::Status::StreamMasterErrorCode::streammaster_error,
+     "streammaster_error"}};
+
+static const std::map<const int, std::string> streamer_error_lookup_{
```
michele-brambilla: That was meant to be a generic error, at the moment is not used. I will get rid of it for now, if at any point will be useful can be restored.
 :white_check_mark:

src/utils.h
michele-brambilla: :white_check_mark:

src/StatusWriter.cxx
```diff
@@ -0,0 +1,144 @@
...
+  auto &a = d.GetAllocator();
+  d.SetObject();
+  { // message type
+    d.AddMember("type", "filewriter_streammaster_status", a);
+  }
+  { // stream master info
+    Value sm,ss;
+    sm.SetObject();
+    std::string s{Err2Str(StreamMasterError{data.status})};
+    ss.SetString(s.c_str(),s.size(),a);
```
michele-brambilla: Workaround: I enabled ``RAPIDJSON_HAS_STDSTRING`` in the headerfile in order to use std::string directly

src/StatusWriter.cxx
```diff
@@ -0,0 +1,144 @@
...
+    Value freq;
+    freq.SetObject();
+    freq.AddMember("avg", x.freq_avg, a);
+    freq.AddMember("std", x.freq_std, a);
+    value.AddMember("freq", freq, a);
+  }
+  return value;
+}
+
+std::string JSONStreamWriter::write(const StreamMasterStatus &data) {
```
michele-brambilla: Actually, virtual shouldn't work, the return types are different. Hiding the base method is exactly what I want. The alternative Dominik suggested me is to have a third class that will be the base for both ``JSONWriter`` and ``JSONStreamWriter``. I will follow his suggestion if it's ok with you

dominikwerder: Or, if code sharing is the goal, then do not share code via inheritance but via composition.

src/Streamer.cxx
michele-brambilla: If the ``connect`` thread for any reason takes longer to be started than the ``writer`` thread, the lock is acquired by the latter, which tries to consume messages before the connection is established. As we discussed the sleep is not 100% safe (nor elegant), I'll try some other solution(a future?)

src/StatusWriter.cxx
matthew-d-jones: After a quick look it seems you want to share `write_impl` and `to_json`, I agree that composition would be most appropriate.

src/tests/status_test.cxx
michele-brambilla: :white_check_mark:

src/Streamer.cxx
matthew-d-jones: A `std::condition_variable` might be appropriate. Whatever creates the threads could pass the `condition_variable` to the `connect` thread and then `wait` on it. The `connect` thread would call `condition_variable.notify` when it has successfully connected, at which point the `wait` terminates and the `writer` thread can be started.

matthew-d-jones: The `wait` timing out gives whatever launches that thread the information that connecting is taking too long, that could also be advantageous.

README.md
michele-brambilla: :white_check_mark:

src/Status.hpp
michele-brambilla: :white_check_mark:

src/StatusWriter.hpp
michele-brambilla: :white_check_mark: (these classes have been removed and replaced)

src/StreamMaster.hpp
michele-brambilla: :white_check_mark:

src/StreamMaster.hpp
michele-brambilla: :white_check_mark:

src/StreamMaster.hpp
michele-brambilla: :white_check_mark: 
removed

src/Streamer.cxx
michele-brambilla: :white_check_mark:

src/Status.cxx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+//  \author Michele Brambilla <michele.brambilla@psi.ch>
+//  \date Wed Oct 11 14:47:17 2017
+
+// data-race safe += with double
+template <class T> void atomic_add(std::atomic<T> &value, const T &other) {
+  auto current = value.load();
+  while (!value.compare_exchange_weak(current, current + other))
+    ;
```
dominikwerder: Isn't this a potential dead-lock? because `value` is never reloaded on failed CAS.

`atomic_fetch_add` would probably suffice?

src/StatusWriter.cxx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+template <class Allocator>
+rapidjson::Value create_derived_quantity(MessageInfo::value_type &value,
+                                         Allocator &a) {
+  rapidjson::Value result;
+  result.SetObject();
+  result.AddMember("average", value.first, a);
+  result.AddMember("stdandard_deviation", value.second, a);
+  return std::move(result);
```
dominikwerder: Why `move` here?

src/StatusWriter.cxx
```diff
@@ -0,0 +1,109 @@
...
+
+JSONStreamWriter::return_type
+JSONStreamWriter::write(StreamMasterInfo &info) const {
+  auto value = base.write_impl(info);
+  rapidjson::StringBuffer buffer;
+  rapidjson::PrettyWriter<rapidjson::StringBuffer> w(buffer);
+  w.SetMaxDecimalPlaces(1);
+  value.Accept(w);
+  std::string s{buffer.GetString()};
+  return std::move(s);
```
dominikwerder: Why `move`?

src/Status.cxx
```diff
@@ -0,0 +1,154 @@
...
+
+#include <rapidjson/document.h>
+
+#include "Status.hpp"
+
+double average(const double &sum, const double &N) { return sum / N; }
+
+double standard_deviation(const double &sum, const double &sum_squared,
+                          const double &N) {
+  auto variance = (N * sum_squared - sum * sum) / (N * (N - 1));
```
dominikwerder: Is there a reason for the additional `N` in numerator and denominator?

src/Status.cxx
michele-brambilla: Unfortunately ```atomic_fetch_add``` wants integers. Indeed you're right, current has to be reloaded upon failure.

src/StatusWriter.cxx
michele-brambilla: :white_check_mark:

src/StatusWriter.cxx
michele-brambilla: :white_check_mark:

src/Status.cxx
michele-brambilla: Just traded one division for two multiplication. I was trying to understand if one of the two solution can be more stable when ```sum_squared``` \approx ```sum * sum```, but found the two equivalent. I can switch to the other form if it looks more clear

src/Status.cxx
```diff
@@ -0,0 +1,152 @@
...
+
+#include <rapidjson/document.h>
+
+#include "Status.hpp"
+
+double average(const double &sum, const double &N) { return sum / N; }
+
+double standard_deviation(const double &sum, const double &sum_squared,
+                          const double &N) {
+  auto variance = (sum_squared - (sum * sum) / N) / (N - 1);
```
dominikwerder: I stumbled again over this formula. I asked in my previous review because I wasn't sure and didn't check myself. But now, when I write it down, I think what you had previously was correct, if the function parameters mean what I think they mean:
`sum_squared = \sum_i x_i^2` and `sum = \sum_i x_i`

src/Report.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <memory>
+#include <thread>
+
+#include "KafkaW.h"
+#include "Status.hpp"
+#include "StatusWriter.hpp"
+#include "logger.h"
+
+namespace KafkaW {
+class ProducerTopic;
```
dominikwerder: Why is the forward declaration needed if you `#include "KafkaW.h"`?

src/Streamer.cxx
```diff
@@ -84,225 +80,251 @@ int FileWriter::Streamer::get_topic_partitions(const std::string &topic) {
...
+FileWriter::Streamer::~Streamer() { close_stream(); }
+
+void FileWriter::Streamer::connect(
+    const std::string topic_name, FileWriter::Streamer::Options kafka_options,
+    FileWriter::Streamer::Options filewriter_options) {
+  std::lock_guard<std::mutex> lock(guard_);
+
+  std::lock_guard<std::mutex> lk(connection_lock_);
+  connection_ready_.notify_all();
+  ready_ = true;
```
dominikwerder: I'm not sure what the `ready_` flag is for, just wanted to ask if it should be maybe rather set before `notify_all()`?

src/Status.cxx
michele-brambilla: As we discussed yesterday yes, ``sum_squared = \sum_i (x_i^2)`` and ``sum = \sum_i x_i``

src/Report.hpp
michele-brambilla: :white_check_mark:

src/Streamer.cxx
michele-brambilla: :white_check_mark:
I also changed their names in order to be at least a little descriptive

PR #22, opened 2017-08-02
Consolidate structures that hold configuration data

dominikwerder:
Currently, configuration options are stored in too many different places.
This PR collects them in MainOpt for easier access.

### DIFF
--- a/src/CommandHandler.cxx
+++ b/src/CommandHandler.cxx
@@ -7,12 +7,12 @@ namespace FileWriter {
 // In the future, want to handle many, but not right now.
 static int g_N_HANDLED = 0;
 
-CommandHandler::CommandHandler(Master *master,
-                               rapidjson::Value const *config_file)
-    : master(master), config_file(config_file) {
+CommandHandler::CommandHandler(MainOpt &config, Master *master)
+    : config(config), master(master) {
+  // Will take care of this in upcoming PR.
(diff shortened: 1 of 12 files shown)


src/CommandHandler.cxx
```diff
@@ -7,12 +7,12 @@ namespace FileWriter {
 // In the future, want to handle many, but not right now.
 static int g_N_HANDLED = 0;
 
-CommandHandler::CommandHandler(Master *master,
-                               rapidjson::Value const *config_file)
-    : master(master), config_file(config_file) {
+CommandHandler::CommandHandler(MainOpt &config, Master *master)
+    : config(config), master(master) {
+  // Will take care of this in upcoming PR.
```
matthew-d-jones: Consider making issues in github rather than leaving comments in the code. Then it is documented in a way which can be referenced in the PR and is harder to forget about.

src/CommandHandler.cxx
```diff
@@ -27,6 +27,7 @@ void CommandHandler::handle_new(rapidjson::Document const &d) {
   // if (g_N_HANDLED > 0) return;
   using namespace rapidjson;
   using std::move;
+  using std::string;
```
matthew-d-jones: Not worth it for two uses of `string`, on one line, in my opinion.

Issue #23, opened 2017-08-04
Use std::regex instead of PCRE


dominikwerder:
After agreement on minimum compiler/library version, use std::regex instead of PCRE.

matthew-d-jones:
Closed by #24

PR #24, opened 2017-08-04
Use std::regex instead of PCRE

dominikwerder:
Issue https://github.com/ess-dmsc/kafka-to-nexus/issues/23


dominikwerder:
Right, thanks!

PR #27, opened 2017-08-04
Publish basic status of file writer `Master`

dominikwerder:
- Let the user define a `status-uri` either via the command line or the configuration file to specify on which Kafka topic the file writer should publish the status
- Publish the basic list of files which are currently being written
- Prepared to pass the shared status producer to `StreamMaster` which in turn will publish more details about each file being written
- The overall status of each file being written will be contained in the published json after DM-450 lands

### DIFF
--- a/src/FileWriterTask.cxx
+++ b/src/FileWriterTask.cxx
@@ -77,11 +77,6 @@ int FileWriterTask::hdf_init(rapidjson::Value const &nexus_structure) {
   return 0;
 }
 
-void FileWriterTask::file_flush() {
-  if (impl)
-    impl->hdf_file.flush();
-}
-
 uint64_t FileWriterTask::id() const { return _id; }
 
 std::string const &FileWriterTask::hdf_filename() const {
--- a/src/FileWriterTask.cxx
+++ b/src/FileWriterTask.cxx
@@ -2,11 +2,15 @@
 #include "HDFFile.h"
 #include "Source.h"
 #include "logger.h"
+#include <atomic>
+#include <chrono>
 
 namespace FileWriter {
 
 using std::string;
 using std::vector;
+
+std::atomic<uint32_t> n_FileWriterTask_created{0};
 
 class FileWriterTask_impl {
   friend class FileWriterTask;
@@ -28,7 +32,14 @@ Source SourceFactory_by_FileWriterTask::create(string topic,
 
 std::vector<DemuxTopic> &FileWriterTask::demuxers() { return _demuxers; }
 
-FileWriterTask::FileWriterTask() { impl.reset(new FileWriterTask_impl); }
+FileWriterTask::FileWriterTask() {
+  using namespace std::chrono;
+  _id = duration_cast<nanoseconds>(system_clock::now().time_since_epoch())
+            .count();
+  _id = (_id & uint64_t(-1) << 16) | (n_FileWriterTask_created & 0xffff);
+  ++n_FileWriterTask_created;
+  impl.reset(new FileWriterTask_impl);
+}
 
 FileWriterTask::~FileWriterTask() { LOG(9, "~FileWriterTask"); }
 
@@ -69,5 +80,7 @@ void FileWriterTask::file_flush() {
   if (impl)
     impl->hdf_file.flush();
 }
+
+uint64_t FileWriterTask::id() const { return _id; }
 
 } // namespace FileWriter
--- a/src/FileWriterTask.cxx
+++ b/src/FileWriterTask.cxx
@@ -83,4 +83,19 @@ void FileWriterTask::file_flush() {
 
 uint64_t FileWriterTask::id() const { return _id; }
 
+rapidjson::Value FileWriterTask::stats(
+    rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> &a) {
+  using namespace rapidjson;
+  Value js_topics;
+  js_topics.SetObject();
+  for (auto &d : _demuxers) {
+    js_topics.AddMember(Value(d.topic().c_str(), a), Value(0), a);
+  }
+  Value js_fwt;
+  js_fwt.SetObject();
+  js_fwt.AddMember("filename", Value(impl->hdf_filename.c_str(), a), a);
+  js_fwt.AddMember("topics", js_topics, a);
+  return js_fwt;
+}
+
 } // namespace FileWriter
--- a/src/FileWriterTask.cxx
+++ b/src/FileWriterTask.cxx
@@ -84,7 +84,7 @@ void FileWriterTask::file_flush() {
 uint64_t FileWriterTask::id() const { return _id; }
 
 rapidjson::Value FileWriterTask::stats(
-    rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> &a) {
+    rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> &a) const {
   using namespace rapidjson;
   Value js_topics;
   js_topics.SetObject();
--- a/src/FileWriterTask.cxx
+++ b/src/FileWriterTask.cxx
@@ -79,10 +79,6 @@ int FileWriterTask::hdf_init(rapidjson::Value const &nexus_structure) {
 
 uint64_t FileWriterTask::id() const { return _id; }
 
-std::string const &FileWriterTask::hdf_filename() const {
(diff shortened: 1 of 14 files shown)


src/FileWriterTask.cxx
```diff
@@ -29,7 +32,14 @@ Source SourceFactory_by_FileWriterTask::create(string topic,
 
 std::vector<DemuxTopic> &FileWriterTask::demuxers() { return _demuxers; }
 
-FileWriterTask::FileWriterTask() { impl.reset(new FileWriterTask_impl); }
+FileWriterTask::FileWriterTask() {
+  using namespace std::chrono;
+  _id = duration_cast<nanoseconds>(system_clock::now().time_since_epoch())
```
matthew-d-jones: You are casting `long` to `uint64_t` here, it could be made explicit with a `static_cast`.

src/FileWriterTask.cxx
```diff
@@ -71,4 +81,25 @@ void FileWriterTask::file_flush() {
     impl->hdf_file.flush();
 }
 
+uint64_t FileWriterTask::id() const { return _id; }
+
+std::string const &FileWriterTask::hdf_filename() const {
```
matthew-d-jones: I don't think this method is ever used

src/FileWriterTask.h
```diff
@@ -30,13 +30,18 @@ class FileWriterTask final {
   std::vector<DemuxTopic> &demuxers();
   void file_flush();
```
matthew-d-jones: `file_flush` is never used
`file_close` isn't implemented

src/MainOpt.cxx
```diff
@@ -63,7 +78,10 @@ std::pair<int, std::unique_ptr<MainOpt>> parse_opt(int argc, char **argv) {
   static struct option long_options[] = {
       {"help", no_argument, 0, 'h'},
       {"config-file", required_argument, 0, 0},
+      {"command-uri", required_argument, 0, 0},
+      // Legacy alias for 'command-uri'
```
matthew-d-jones: Let's try to avoid taking on legacy baggage already, do you have a really good reason for not breaking backwards compatibility at this point?

src/MainOpt.cxx
```diff
@@ -63,7 +78,10 @@ std::pair<int, std::unique_ptr<MainOpt>> parse_opt(int argc, char **argv) {
   static struct option long_options[] = {
       {"help", no_argument, 0, 'h'},
       {"config-file", required_argument, 0, 0},
+      {"command-uri", required_argument, 0, 0},
```
matthew-d-jones: The second to last argument is a pointer, use `nullptr` rather than `0`

src/MainOpt.h
```diff
@@ -43,6 +43,14 @@ struct MainOpt {
   /// Kafka broker and topic where file writer commands are published.
   uri::URI command_broker_uri{"kafka://localhost:9092/kafka-to-nexus.command"};
 
+  /// Whether we want to publish status to Kafka
+  bool do_kafka_status = false;
```
matthew-d-jones: I like the explicit variable naming and doc comments here :)

src/Master.cxx
```diff
@@ -2,11 +2,13 @@
 #include "CommandHandler.h"
 #include "FileWriterTask.h"
 #include "Source.h"
+#include "commandproducer.h"
 #include "logger.h"
 #include <chrono>
-// #include "helper.h"
-// #include "utils.h"
-#include "commandproducer.h"
+#include <rapidjson/document.h>
```
matthew-d-jones: It looks like these includes are unused:
```
#include "FileWriterTask.h"
#include "Source.h"
#include "logger.h"
#include <chrono>
#include <rapidjson/document.h>
#include <rapidjson/prettywriter.h>
#include <sys/types.h>
```

src/Master.cxx
```diff
@@ -15,7 +17,19 @@ using std::string;
...
 std::string &CmdMsg_K::str() { return _str; }
 
-Master::Master(MainOpt &config) : config(config), command_listener(config) {}
+Master::Master(MainOpt &config) : config(config), command_listener(config) {
+  std::vector<char> buffer;
+  buffer.resize(128);
+  gethostname(buffer.data(), buffer.size());
+  if (buffer.back() != 0) {
+    // likely an error
+    buffer.back() = 0;
```
matthew-d-jones: I think you want to `push_back` here rather than overwrite the last character?

src/Master.cxx
```diff
@@ -28,17 +42,33 @@ void Master::handle_command(rapidjson::Document const &cmd) {
...
+  using Clock = std::chrono::steady_clock;
+  auto t_last_statistics = Clock::now();
   while (do_run) {
     LOG(7, "Master poll");
     auto p = command_listener.poll();
     if (auto msg = p.is_Msg()) {
       LOG(7, "Handle a command");
       this->handle_command_message(std::move(msg));
     }
+    if (Clock::now() - t_last_statistics >
```
matthew-d-jones: Should this always happen, or only when `config.do_kafka_status`?

src/Master.cxx
```diff
@@ -47,6 +77,40 @@ void Master::run() {
...
+    js_files.AddMember(Value(fwt_id_str.c_str(), a), fwt, a);
+  }
+  js_status.AddMember("files", js_files, a);
+  rapidjson::StringBuffer buffer;
+  rapidjson::PrettyWriter<rapidjson::StringBuffer> writer(buffer);
+  writer.SetIndent(' ', 2);
+  js_status.Accept(writer);
+  if (status_producer) {
+    status_producer->produce((KafkaW::uchar *)buffer.GetString(),
+                             buffer.GetSize());
```
matthew-d-jones: This casts `size_t` to `int`, make it explicit with a `static_cast` if it is intentional.

src/kafka-to-nexus.cxx
```diff
@@ -33,12 +33,19 @@ int main(int argc, char **argv) {
            "\n"
            "  --config-file               <filename.json>\n"
            "\n"
-           "  --broker-command            <//host[:port][/topic]>\n"
-           "      Kafka brokers to connect with for configuration updates.\n"
+           "  --command-uri               <//host[:port][/topic]>\n"
```
matthew-d-jones: These changes look good, but `std::cout` should be used instead of `printf` in C++

src/FileWriterTask.cxx
dominikwerder: It is in [src/Master.cxx:89](https://github.com/ess-dmsc/kafka-to-nexus/blob/master_status/src/Master.cxx#L89)

src/MainOpt.cxx
dominikwerder: No, I don't.

src/FileWriterTask.cxx
matthew-d-jones: `hdf_filename()` not `id()` :)

src/Master.cxx
dominikwerder: What is our policy?
Some projects include the headers which are necessary for the explicitly used types in the source file even if they are already pulled as dependency from another include.
So I use for example explicitly std::chrono, therefore I include <chrono> even though it would be already pulled in through another header.

src/Master.cxx
dominikwerder: The buffer size is chosen such that every possible hostname should fit in. Problem is that it is not trivial to find the required size of the buffer. If for some reason it didn't fit, then there is anyway no way to find out whether only the null terminator got left out. So we can as well just overwrite the last byte. This is really just for safety. I can add a runtime assert and log...

src/Master.cxx
matthew-d-jones: Interesting, I haven't come across such a policy. It is difficult to distinguish between headers which are unused because they are included elsewhere or unused because they are completely unneeded, therefore I'd rather remove them. Do you prefer to keep them?

src/Master.cxx
matthew-d-jones: Ah I see, perhaps just log a "warning" when this happens?

src/Master.cxx
dominikwerder: Right, I should get rid of this int.

src/FileWriterTask.cxx
dominikwerder: Oh, that's right!

PR #33, opened 2017-08-10
JSON helper function merge() to merge two rapidjson structures

dominikwerder:
With tests.
To be used in upcoming feature branch.

### DIFF
--- a/src/json.cxx
+++ b/src/json.cxx
@@ -57,37 +57,33 @@ rapidjson::Document merge(rapidjson::Value const &v1,
   while (stack.size() > 0) {
     auto &se = stack.back();
     if (se.it1 != se.v1->MemberEnd()) {
-      // Naming scheme:
-      // m for 'member'
-      // (1|2) if it is out of 1st or 2nd input document.
-      // (n|v) name or value.
-      auto &m1n = se.it1->name;
-      auto &m1v = se.it1->value;
+      auto &first_doc_member_name = se.it1->name;
+      auto &first_doc_member_value = se.it1->value;
       // Let's see if the 2nd input document has a member of the same name. If
       // not, there is nothing to do.
       auto it2 = se.v2->FindMember(se.it1->name.GetString());
       if (it2 != se.v2->MemberEnd()) {
-        auto &m2v = it2->value;
-        if (m1v.IsObject()) {
-          if (m2v.IsObject()) {
+        auto &second_doc_member_value = it2->value;
+        if (first_doc_member_value.IsObject()) {
+          if (second_doc_member_value.IsObject()) {
             // If both 1st and 2nd input are json objects at this level then we
             // recursively continue the comparison.
-            stack_element e{m1v.MemberBegin(),
-                            &se.v1->operator[](m1n.GetString()),
-                            &m2v,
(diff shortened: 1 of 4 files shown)


src/json.cxx
```diff
@@ -15,3 +16,69 @@ std::string json_to_string(rapidjson::Document const &jd) {
...
+  struct stack_element {
+    Value::ConstMemberIterator it1;
+    Value *v1;
+    Value const *v2;
+    deque<to_copy_element> to_copy;
+  };
+  deque<stack_element> stack;
+  stack.push_back(stack_element{ret.MemberBegin(), &ret, &v2});
+
+  while (stack.size() > 0) {
```
matthew-d-jones: The `while` loop is not very easy to read. Could you extract some methods with descriptive method names, for example for the `for` loops at line 69 and 74? Another good option would be adding a few comments to explain key lines.

src/json.cxx
```diff
@@ -15,3 +16,69 @@ std::string json_to_string(rapidjson::Document const &jd) {
...
+  auto &a = ret.GetAllocator();
+  ret.CopyFrom(v1, a);
+  if (!v1.IsObject() || !v2.IsObject()) {
+    return ret;
+  }
+  struct to_copy_element {
+    string name;
+    Value const *v;
+  };
+  struct stack_element {
```
matthew-d-jones: This looks good to me. I think you've made the right choice of container (`deque`) and constraining these structs to the scope of the method.

src/tests/json.cxx
```diff
@@ -0,0 +1,105 @@
+#include "../json.h"
+#include "../logger.h"
+#include <gtest/gtest.h>
+
+using namespace rapidjson;
+
+TEST(json, merge_01) {
```
matthew-d-jones: I advise against this style of numbered unit test names. This is because explicit names for what you are testing make it easy for someone else to see what the code does. The test itself ensures that the code does what you intend, and the descriptive name makes it easy for someone else to spot the other possibility of a bug: that the intent of the programmer is flawed.
This test, for example, could be called `check_element_in_second_document_replaces_element_in_first_document_with_same_key`

dominikwerder: :+1:

src/json.cxx
dominikwerder: The rapidjson api tends to be a bit on the verbose side, but I decided to add comments instead of taking out the loop.

src/json.cxx
```diff
@@ -15,3 +16,103 @@ std::string json_to_string(rapidjson::Document const &jd) {
...
+  deque<stack_element> stack;
+  stack.push_back(stack_element{ret.MemberBegin(), &ret, &v2});
+
+  while (stack.size() > 0) {
+    auto &se = stack.back();
+    if (se.it1 != se.v1->MemberEnd()) {
+      // Naming scheme:
+      // m for 'member'
+      // (1|2) if it is out of 1st or 2nd input document.
+      // (n|v) name or value.
```
matthew-d-jones: :+1: The comments below are very helpful. 
Variable names should not require a key though. How about using `first_doc_member_name`, `second_doc_member_value` etc?

dominikwerder: ✔

PR #34, opened 2017-08-17
Refactor the local HDF file writer test case

dominikwerder:
Refactor the `data_ev42` local HDF file writer test case to allow for better profiling and to have a better basis for changing the json command format in an upcoming PR.

### DIFF
--- a/src/h5.cxx
+++ b/src/h5.cxx
@@ -333,7 +333,9 @@ append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
     }
     buf_n += nbytes;
   }
-  if (buf_n >= buf_SIZE - 2 * buf_MAXPKG || (!do_buf && buf_n > 0)) {
+  // Flush the buffer if there is a chance that it will be full on the next
+  // iteration.
+  if (buf_n > buf_SIZE - buf_MAXPKG || (!do_buf && buf_n > 0)) {
     if (flush_buf() != 0) {
       return {-1};
     }
--- a/src/h5.cxx
+++ b/src/h5.cxx
@@ -304,7 +304,9 @@ h5d_chunked_1d<T>::create(hid_t loc, string name, hsize_t chunk_bytes) {
 template <typename T>
 h5d_chunked_1d<T>::h5d_chunked_1d(hid_t loc, string name, hsize_t chunk_bytes,
                                   h5d ds)
-    : ds(move(ds)), dsp_wr(this->ds) {}
+    : ds(move(ds)), dsp_wr(this->ds) {
+  buf.resize(buf_SIZE);
+}
 
 template <typename T>
 h5d_chunked_1d<T>::h5d_chunked_1d(h5d_chunked_1d &&x)
@@ -321,11 +323,13 @@ template <typename T> void swap(h5d_chunked_1d<T> &x, h5d_chunked_1d<T> &y) {
 template <typename T>
 append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
   append_ret ret{-1};
-  bool do_buf = nlen * sizeof(T) < 4 * 1024;
+  auto nbytes = nlen * sizeof(T);
+  bool do_buf = nbytes <= buf_MAXPKG;
   if (do_buf) {
-    std::copy(data, data + nlen, std::back_inserter(buf));
+    memcpy(buf.data() + buf_n, data, nbytes);
+    buf_n += nbytes;
   }
-  if (buf.size() > 128 * 1024 || (!do_buf && buf.size() > 0)) {
+  if (buf_n >= buf_SIZE - 2 * buf_MAXPKG || (!do_buf && buf_n > 0)) {
     if (flush_buf() != 0) {
       return {-1};
     }
@@ -338,17 +342,17 @@ append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
   }
   ret.status = 0;
   ret.ix0 = i0;
-  ret.written_bytes = sizeof(T) * nlen;
+  ret.written_bytes = nbytes;
   i0 += nlen;
   return ret;
 }
 
 template <typename T> int h5d_chunked_1d<T>::flush_buf() {
-  auto wr = ds.append_data_1d(buf.data(), buf.size());
+  auto wr = ds.append_data_1d((T *)buf.data(), buf_n / sizeof(T));
   if (!wr) {
     return -1;
   }
-  buf.clear();
+  buf_n = 0;
   return 0;
 }
 
--- a/src/h5.cxx
+++ b/src/h5.cxx
@@ -323,12 +323,12 @@ template <typename T> void swap(h5d_chunked_1d<T> &x, h5d_chunked_1d<T> &y) {
 template <typename T>
 append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
   append_ret ret{-1};
-  uint32_t nbytes = nlen * sizeof(T);
+  auto nbytes = nlen * sizeof(T);
   bool do_buf = nbytes <= buf_MAXPKG;
   if (do_buf) {
     auto p1 = (char *)data;
     auto p2 = buf.data() + buf_n;
-    for (uint32_t i1 = 0; i1 < nbytes; ++i1) {
+    for (size_t i1 = 0; i1 < nbytes; ++i1) {
       p2[i1] = p1[i1];
     }
     buf_n += nbytes;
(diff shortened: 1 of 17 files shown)


src/h5.cxx
```diff
@@ -321,11 +323,17 @@ template <typename T> void swap(h5d_chunked_1d<T> &x, h5d_chunked_1d<T> &y) {
...
-    std::copy(data, data + nlen, std::back_inserter(buf));
+    auto p1 = (char *)data;
+    auto p2 = buf.data() + buf_n;
+    for (uint32_t i1 = 0; i1 < nbytes; ++i1) {
+      p2[i1] = p1[i1];
+    }
+    buf_n += nbytes;
   }
-  if (buf.size() > 128 * 1024 || (!do_buf && buf.size() > 0)) {
+  if (buf_n >= buf_SIZE - 2 * buf_MAXPKG || (!do_buf && buf_n > 0)) {
```
matthew-d-jones: What's the significance of `buf_SIZE - 2 * buf_MAXPKG`? Could you add a comment to clarify?

src/h5.h
```diff
@@ -94,7 +94,10 @@ template <typename T> class h5d_chunked_1d {
   h5d_chunked_1d();
```
matthew-d-jones: I know it isn't in your changes, but this constructor is not actually implemented.

src/schemas/ev42/ev42_rw.cxx
```diff
@@ -83,26 +83,34 @@ void writer::init_impl(std::string const &sourcename, hid_t hdf_group,
                        Msg msg) {
   LOG(7, "ev42::init_impl");
 
+  hsize_t chunk_n_elements = 1;
+
   if (config_file) {
     if (auto x = get_int(config_file, "nexus.indices.index_every_kb")) {
       index_every_bytes = (int)x * 1024;
```
matthew-d-jones: There is a signed to unsigned cast here which should be explicit if intentional, and the C-style could be removed in the process.

src/schemas/ev42/ev42_rw.cxx
```diff
@@ -83,26 +83,34 @@ void writer::init_impl(std::string const &sourcename, hid_t hdf_group,
...
+      LOG(7, "index_every_bytes: {}", index_every_bytes);
     } else if (auto x = get_int(config_file, "nexus.indices.index_every_mb")) {
       index_every_bytes = (int)x * 1024 * 1024;
```
matthew-d-jones: Same as previous comment

src/schemas/ev42/ev42_rw.cxx
```diff
@@ -83,26 +83,34 @@ void writer::init_impl(std::string const &sourcename, hid_t hdf_group,
...
+      LOG(7, "index_every_bytes: {}", index_every_bytes);
+    }
+    if (auto x = get_int(config_file, "nexus.chunk.chunk_n_elements")) {
+      chunk_n_elements = x.v;
```
matthew-d-jones: Casting `int` to `hsize_t`, make explicit if intentional.

src/tests/HDFFile.cxx
```diff
@@ -7,10 +7,33 @@
 #include "../helper.h"
```
matthew-d-jones: This file has the same name as the implementation it is testing, consider adding "Test" to the name.

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
+    std::mt19937 rnd;
+    vector<FlatBufs::ev42::fb> fbs;
+    vector<FileWriter::Msg> msgs;
+    // Number of messages already fed into file writer during testing
+    size_t n_fed = 0;
+    /// Generates n test messages which we can later feed from memory into the
+    /// file writer.
+    void pregenerate(int n, int n_events_per_message) {
+      // LOG(4, "generating {} {}...", topic, source);
+      FlatBufs::ev42::synth synth(source, seed);
```
matthew-d-jones: `seed` is being cast to `uint64_t` here and in the line below.

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
+      rnd.seed(seed);
+      for (int i1 = 0; i1 < n; ++i1) {
+        // Number of events per message:
+        // size_t n_ele = rnd() >> 24;
+        // Currently fixed, have to adapt verification code first.
+        size_t n_ele = n_events_per_message;
```
matthew-d-jones: Casting signed `int` to `size_t` here and then back to `int` on line 167.

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
+    int seed = 0;
+    std::mt19937 rnd;
+    vector<FlatBufs::ev42::fb> fbs;
+    vector<FileWriter::Msg> msgs;
+    // Number of messages already fed into file writer during testing
+    size_t n_fed = 0;
+    /// Generates n test messages which we can later feed from memory into the
+    /// file writer.
+    void pregenerate(int n, int n_events_per_message) {
+      // LOG(4, "generating {} {}...", topic, source);
```
matthew-d-jones: Why the commented log here and lines 171, 250? Should they be uncommented or removed?

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
 
-    auto dt = H5Dget_type(ds);
-    ASSERT_GE(dt, 0);
-    ASSERT_EQ(H5Tget_size(dt), sizeof(DT));
+    for (int file_i = 0; file_i < 1; ++file_i) {
+      unlink(string(fname).c_str());
+
+      Msg msg;
+      msg.data = (char *)cmd.data();
+      msg.size = cmd.size();
```
matthew-d-jones: I think this is preferable to having the uninitialised type and then populating in parts:
```
Msg msg = {
    (char *)cmd.data(),
    cmd.size(),
};
```

matthew-d-jones: You are also casting an `unsigned long` to `int32_t` here. I suspect `Msg::size` being an `int32_t` is a mistake.

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
+      using MS = std::chrono::milliseconds;
+      auto t1 = CLK::now();
+      for (;;) {
+        bool change = false;
+        for (int i1 = 0; i1 < feed_msgs_times; ++i1) {
+          for (auto &source : sources) {
+            for (int i2 = 0;
+                 i2 < n_msgs_per_batch && source.n_fed < source.msgs.size();
+                 ++i2) {
+              auto &msg = source.msgs[source.n_fed];
```
matthew-d-jones: This hides the `msg` declared in the upper scope, but refactoring this block to a separate method would solve this.

src/tests/HDFFile.cxx
```diff
@@ -118,155 +141,323 @@ class T_CommandHandler : public testing::Test {
...
+        fbs.push_back(synth.next(n_ele));
+        auto &fb = fbs.back();
+        msgs.push_back(FileWriter::Msg{(char *)fb.builder->GetBufferPointer(),
+                                       (int32_t)fb.builder->GetSize()});
+        // LOG(4, "GetSize: {}", fb.builder->GetSize());
+      }
+    }
+  };
+
   static void data_ev42() {
```
matthew-d-jones: This method is huge. I think it would be a _lot_ easier to follow if some separate methods, with clear names, were extracted from it. That would also solve a couple of places where you reuse variable names.

src/tests/HDFFile.cxx
dominikwerder: Do you mean the `HDFFile`? Test code has the same name, but in directory `tests`.

src/tests/HDFFile.cxx
dominikwerder: True.  This part has already changed quite a bit though in another upcoming feature branch, so I'd like to skip to break it up in this PR to avoid the rebase.

dominikwerder:
Thank you for your review, I've pushed my changes.

src/tests/HDFFile.cxx
matthew-d-jones: Yes, I mean you have two files called `HDFFile.cxx`. I think that can lead to confusion, particularly as it is usual to navigate around a codebase by typing file or class names.

dominikwerder: Is there really a clear-cut answer?  Both approaches have pros and cons.  Let's discuss, then document the policy somewhere.

PR #37, opened 2017-09-13
Headers in cmake file

SkyToGround:
I have modified the CMakeLists.txt file to define the actual header files of interest. This simplifies development significantly when generating IDE project files (e.g. Xcode, Eclipse) from the CMake files.

I have also made the binary output paths less ambiguous and I have added indentations where appropriate.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,10 +1,10 @@
 # Relative to build root
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ..)
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ..)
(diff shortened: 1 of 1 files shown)


matthew-d-jones:
@SkyToGround Could you expand on "simplifies development", the Eclipse and Xcode generators shouldn't require explicit header file names?

SkyToGround:
Sure. When generating IDE (E.g. Xcode, Eclipse, MSV etc.) project files from CMake-files, the project files contains links (or references) to the files defined in those CMake files. If the headers are not defined, the projects does not contain references to the header files (i.e. you loose code completion functionality etc.).

SkyToGround:
As one of many references, see the following stackoverflow post: https://stackoverflow.com/questions/13703647/how-to-properly-add-include-directories-with-cmake

PR #43, opened 2017-10-10
Conan

michele-brambilla:
Add Conan support for getting dependencies and building project. Instructions are provided in the README.


Jenkinsfile
```diff
@@ -1,56 +1,85 @@
...
 
-    dir("build") {
-        stage("Update local dependencies") {
-            sh "cd .. && bash code/build-script/update-local-deps.sh"
+    try {
+        container = centos.run(run_args)
+
+	cleanWs()
+
+        stage('Checkout') {
```
amues: We've recently changed the checkout stage for other jobs using Conan and Docker for keeping the change information on Jenkins. This can be done by cloning the repository to the node and then copying it to the container by running
```
// Outside container
dir("${project}") {
    stage('Checkout') {
        scm_vars = checkout scm
    }
}

...

// Inside container
// Copy sources to container and change owner and group.
sh "docker cp ${project} ${container_name}:/home/jenkins/${project}"
sh """docker exec --user root ${container_name} ${custom_sh} -c \"
    chown -R jenkins.jenkins /home/jenkins/${project}
\""""
```

src/CMakeLists.txt
```diff
@@ -57,11 +52,12 @@ set(libraries_common
     ${CURL_LIBRARIES}
     pthread
     z
+    ${CONAN_LIBS}
```
amues: Did the build work with the removed `find_packages` (lines 11-17 above)? Leaving them in the CMakeLists.txt file, without including `${CONAN_LIBS}` allows the project to build with and without Conan, whereas using this variable couples the build to it. If it works with the generic find_packages, I would favour keeping them instead.

michele-brambilla: You're absolutely right. I already restored the ``find_packages`` but not yet pushed this change. Thank you for pointing out

dominikwerder: :+1: can confirm that this currently breaks non-Conan builds.
Out of curiosity: Why did the corresponding `conan` branch in the `forward-epics-to-kafka` not require `$CONAN_LIBS`?

amues: If the `find_packages` commands don't assume any specific location for the packages, it should work with Conan without `$CONAN_LIBS`. That was the case for the Forwarder. If you encounter a problem with this, please let me know.

Jenkinsfile
```diff
@@ -1,56 +1,97 @@
...
+	cleanWs()
+
+#        stage('Checkout') {
+#            def checkout_script = """
+#                git clone https://github.com/ess-dmsc/${project}.git \
+#                    --branch ${env.BRANCH_NAME}
+#                git clone -b master https://github.com/ess-dmsc/streaming-data-types.git
+#            """
+#            sh "docker exec ${container_name} ${sclsh} -c \"${checkout_script}\""
+#        }
```
amues: This broke the build – comments in Groovy use `//` 😀

amues:
Thanks for the Conan support addition!

michele-brambilla:
Thanks for your support in the integration!

PR #52, opened 2017-11-13
Do not require type for neutron event data

dominikwerder:
The NXevent_data already specifies the types so there is no need to require a
type in the filewriter write command.


src/schemas/ev42/ev42_rw.cxx
```diff
@@ -93,13 +93,6 @@ HDFWriterModule::init_hdf(hid_t hdf_file, std::string hdf_parent_name,
     return HDFWriterModule::InitResult::ERROR_INCOMPLETE_CONFIGURATION();
   }
   auto sourcename = str.v;
```
matthew-d-jones: `sourcename` is now unused, can be removed.

dominikwerder: :+1:

PR #53, opened 2017-11-13
Check hdf dataset create

dominikwerder:
`f142` writer module now cares about a hdf dataset creation failure also in the scalar case.

Also, it now checks that the `value_type()` matches the expectations.

### DIFF
--- a/src/schemas/f142/f142_rw.cxx
+++ b/src/schemas/f142/f142_rw.cxx
@@ -73,7 +73,8 @@ writer_typed_array<DT, FV>::writer_typed_array(hid_t hdf_group,
 
 template <typename DT, typename FV>
 append_ret writer_typed_array<DT, FV>::write_impl(FBUF const *fbuf) {
-  if (fbuf->value_type() != fb_value_type_id) {
+  auto vt = fbuf->value_type();
+  if (vt == Value::NONE || vt != fb_value_type_id) {
     return {-2, 0, 0};
   }
   auto v1 = (FV const *)fbuf->value();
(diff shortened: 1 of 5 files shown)


src/schemas/f142/f142_rw.cxx
```diff
@@ -78,19 +92,31 @@ append_ret writer_typed_array<DT, FV>::write_impl(FBUF const *fbuf) {
 }
 
 template <typename DT, typename FV>
-writer_typed_scalar<DT, FV>::writer_typed_scalar(
-    hid_t hdf_group, std::string const &sourcename) {
+writer_typed_scalar<DT, FV>::writer_typed_scalar(hid_t hdf_group,
+                                                 std::string const &sourcename,
+                                                 Value fb_value_type_id)
+    : fb_value_type_id(fb_value_type_id) {
```
matthew-d-jones: The member variable should not be named the same as the constructor argument. You were using an underscore prefix for member variables in some places, that would solve the problem here.

src/schemas/f142/f142_rw.cxx
```diff
@@ -78,19 +92,31 @@ append_ret writer_typed_array<DT, FV>::write_impl(FBUF const *fbuf) {
 }
 
 template <typename DT, typename FV>
-writer_typed_scalar<DT, FV>::writer_typed_scalar(
-    hid_t hdf_group, std::string const &sourcename) {
+writer_typed_scalar<DT, FV>::writer_typed_scalar(hid_t hdf_group,
+                                                 std::string const &sourcename,
```
matthew-d-jones: `sourcename` ought to be underscore separated: `source_name`.

PR #63, opened 2017-11-23
Return timestamp of written message back to caller

### LINKED ISSUE #62: HDF writer modules do not return back the timestamp
After the change in the HDFWriterModule interface, the writers do no longer return the timestamp of the written message.

dominikwerder:
Closes #62


michele-brambilla:
Works for me, thanks

PR #65, opened 2017-11-29
String datasets

### LINKED ISSUE #57: Dataset with fixed length strings
An option to make the length of strings fixed in the nexus format will be useful.

dominikwerder:
Closes #57 
Better support for writing datasets of type string.

### DIFF
--- a/src/HDFFile.cxx
+++ b/src/HDFFile.cxx
@@ -237,6 +237,45 @@ static void populate_string_pointers(std::vector<char const *> &ptrs, rapidjson:
   }
 }
 
+static void populate_string_fixed_size(std::vector<char> &blob, hsize_t element_size, rapidjson::Value const *vals) {
+  size_t n_added = 0;
+  if (vals->IsString()) {
+    memcpy(blob.data() + n_added * element_size, vals->GetString(), vals->GetStringLength());
+    ++n_added;
+  } else if (vals->IsArray()) {
+    std::vector<rapidjson::Value const *> as;
+    std::vector<size_t> ai;
+    std::vector<size_t> an;
+    as.push_back(vals);
+    ai.push_back(0);
+    an.push_back(vals->GetArray().Size());
+
+    while (not as.empty()) {
+      if (as.size() > 10) {
+        break;
+      }
+      if (ai.back() >= an.back()) {
+        as.pop_back();
+        ai.pop_back();
+        an.pop_back();
+        continue;
+      }
+      auto &v = as.back()->GetArray()[ai.back()];
+      if (v.IsArray()) {
+        ai.back()++;
+        as.push_back(&v);
+        ai.push_back(0);
+        size_t n = v.GetArray().Size();
+        an.push_back(n);
+      } else if (v.IsString()) {
+        memcpy(blob.data() + n_added * element_size, v.GetString(), v.GetStringLength());
+        ++n_added;
+        ai.back()++;
+      }
+    }
+  }
+}
+
 template <typename DT>
 static void
 write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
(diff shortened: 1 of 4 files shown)


src/HDFFile.cxx
```diff
@@ -201,6 +201,81 @@ static void populate_blob(std::vector<DT> &blob, rapidjson::Value const *vals) {
   }
 }
 
+static void populate_string_pointers(std::vector<char const *> &ptrs, rapidjson::Value const *vals) {
+  if (vals->IsString()) {
+    ptrs.push_back(vals->GetString());
+  } else if (vals->IsArray()) {
+    std::vector<rapidjson::Value const *> as;
```
matthew-d-jones: You are using these vectors as stacks, so why not use `std::stack`?

src/HDFFile.cxx
```diff
@@ -201,6 +201,81 @@ static void populate_blob(std::vector<DT> &blob, rapidjson::Value const *vals) {
...
+        ai.back()++;
+      }
+    }
+  }
+}
+
+static void populate_string_fixed_size(std::vector<char> &blob, hsize_t element_size, rapidjson::Value const *vals) {
+  size_t n_added = 0;
+  if (vals->IsString()) {
+    memcpy(blob.data() + n_added * element_size, vals->GetString(), vals->GetStringLength());
```
matthew-d-jones: `std::copy` is preferred over `memcpy`, it is safer (works with non-POD types) and interestingly can be optimised better by the compiler in some situations (because it can make use of the retained type information).
It would be something like this (untested!):
```
auto string_in_json = vals->GetString();
std::copy(string_in_json, string_in_json + vals->GetStringLength(),
          std::back_inserter(blob));
```

src/HDFFile.cxx
```diff
@@ -243,10 +318,104 @@ write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
...
+  for (auto x : sizes) {
+    total_n *= x;
+  }
+  auto dcpl = H5Pcreate(H5P_DATASET_CREATE);
+  hid_t dsp = -1;
+  if (sizes.empty()) {
+    dsp = H5Screate(H5S_SCALAR);
+  } else {
+    dsp = H5Screate(H5S_SIMPLE);
+    H5Sset_extent_simple(dsp, (int)sizes.size(), sizes.data(), max.data());
```
matthew-d-jones: Please use `static_cast<int>` explicitly.

src/HDFFile.cxx
```diff
@@ -243,10 +318,104 @@ write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
...
+    if (max[0] == H5S_UNLIMITED) {
+      H5Pset_chunk(dcpl, sizes.size(), sizes.data());
```
matthew-d-jones: The size should also be explicitly cast here.

src/HDFFile.cxx
```diff
@@ -243,10 +318,104 @@ write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
...
+
+  auto dt = H5Tcopy(H5T_C_S1);
+  H5Tset_size(dt, H5T_VARIABLE);
+  H5Tset_cset(dt, H5T_CSET_UTF8);
+
+  auto ds = H5Dcreate2(hdf_parent, name.data(), dt, dsp, H5P_DEFAULT, dcpl,
+                       H5P_DEFAULT);
+  auto err = H5Dwrite(ds, dt, H5S_ALL, H5S_ALL, H5P_DEFAULT, blob.data());
+  if (err < 0) {
+    LOG(3, "error while writing dataset");
```
matthew-d-jones: Are there different error codes such that you could give more specific error messages, or put the error code in the log message?

src/HDFFile.cxx
```diff
@@ -243,10 +318,104 @@ write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
...
+  }
+  auto dcpl = H5Pcreate(H5P_DATASET_CREATE);
+  hid_t dsp = -1;
+  if (sizes.empty()) {
+    dsp = H5Screate(H5S_SCALAR);
+  } else {
+    dsp = H5Screate(H5S_SIMPLE);
+    H5Sset_extent_simple(dsp, (int)sizes.size(), sizes.data(), max.data());
+    if (max[0] == H5S_UNLIMITED) {
+      H5Pset_chunk(dcpl, sizes.size(), sizes.data());
```
matthew-d-jones: size cast

src/HDFFile.cxx
```diff
@@ -243,10 +318,104 @@ write_ds_numeric(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
...
+
+static void
+write_ds_string_fixed_size(hid_t hdf_parent, std::string name, std::vector<hsize_t> sizes,
+                 std::vector<hsize_t> max, hsize_t element_size, rapidjson::Value const *vals) {
+  size_t total_n = 1;
+  for (auto x : sizes) {
+    total_n *= x;
+  }
+  auto dcpl = H5Pcreate(H5P_DATASET_CREATE);
+  hid_t dsp = -1;
```
matthew-d-jones: This value is never used, could you just declare and not assign?

src/HDFFile.cxx
dominikwerder: The advantages you list do not apply in the special case here, though. The advantage I do see is that `std::copy` is more idiomatic, should that piece of code get reused in the future.

matthew-d-jones:
@dominikwerder Thanks for the changes. Could you make the `std::stack` and `std::copy` changes also to `populate_string_pointers()` and `populate_string_fixed_size()`? Then I'm happy to merge.

dominikwerder:
:+1:

PR #96, opened 2018-01-03
Reduction in log message severity level ambiguity

SkyToGround:
The ambiguity in the log message severity level required some attention and I have for this reason modified the logging code a bit. These are the changes I have made:

* Added severity levels in the form of `enum class Sev`. This class is found in `logger.h` and defines 8 (syslog based) severity levels. The levels are commented in order to help you select which one to use.

* `dwlog()` has been modified to require the severity level in the form of a `enum class Sev` instance.

* All calls to `dwlog()` (the macro `LOG`) have been modified with a `enum class Sev` level to replace the previous `int` level.

* The argument parsers of of *kafka-to-nexus* and *send-command* have been modified to take an explicit severity level (as an integer) instead of as previously, counting the number of occurrences of the `-v` flag.

I tried to infer the appropriate severity levels when replacing the old integer based severity levels.  However, some one more familiar with the code should probably look through all calls to `dwlog()`(`LOG`), there are only 242 of them.

### DIFF
--- a/src/CommandHandler.cxx
+++ b/src/CommandHandler.cxx
@@ -60,7 +60,7 @@ CommandHandler::CommandHandler(MainOpt &config, Master *master)
     auto doc = make_unique<rapidjson::Document>();
     ParseResult err = doc->Parse(buf1.data(), buf1.size());
     if (err.Code() != ParseErrorCode::kParseErrorNone) {
-      LOG(Sev::Err, "ERROR can not parse schema_command");
+      LOG(Sev::Error, "ERROR can not parse schema_command");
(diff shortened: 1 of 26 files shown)


SkyToGround:
@matthew-d-jones I mostly agree with your suggestion. My thinking was that it was easier to get this change accepted if abbreviations were used. But if no-one has any strong objections I will happily change the code to the full names of the severity levels (except Informational).

SkyToGround:
I have now changed the severity level names to their original long versions. Assuming this commit builds properly and that no-one has any objections, I will merge the changes tomorrow (Thursday).

PR #97, opened 2018-01-15
Improve start stop

### LINKED ISSUE #54: The start_time in JSON command not being respected
When starting file writing with some kafka streams attached, I noticed that the data was written before the `start_time`. This would probably mean that `start_time` is not properly configured.

### LINKED ISSUE #55: Stopping the file writer not working properly
When issuing the stop command with or without the `stop_time` does not flush the file until the file writer is closed. From the logs it is evident that the file writer sees a stop has been requested. But it does not flush the file and the data is still being written.

michele-brambilla:
This PR closes #54, closes #55. 

- If the user provides a start time the ``Streamer`` uses ``offsetsForTimes`` to make sure that there are no earlier messages in the log. Indeed this is not enough, as  #54 points out. Usually there are unwanted messages among the offset the method returns and the actual target. The solution is in ``Streamer::write``: the message timestamp is extracted using ``DemuxTopic::time_difference_from_message`` and compared against the start time. Earlier messages are discarded.
- After listening for new commands ``Master`` checks the status off all the ``StreamMaster``. If any of them is in the ``StreamMasterErrorCode::is_removable`` status is removed from the list (erased).
- If the user issues a stop command specifying the stop time ``StreamMaster::forceStop`` makes sure that at most 5 seconds after the stop time (comparing with the system time) all the streams are closed and the ``StreamMaster`` is removable.


src/tests/CMakeLists.txt
```diff
@@ -23,17 +24,19 @@ target_link_libraries(${tgt} ${libraries_common})
...
-streamer_test.cxx
-$<TARGET_OBJECTS:kafka_to_nexus__objects>
-)
-add_executable(${tgt} ${sources})
-add_dependencies(${tgt} flatbuffers_generate)
-target_compile_definitions(${tgt} PRIVATE ${compile_defs_common})
-target_include_directories(${tgt} PRIVATE ${path_include_common})
-target_link_libraries(${tgt} ${libraries_common})
-add_gtest_to_target(${tgt})
+#set(tgt "streamer_test")
```
matthew-d-jones: What is going on here, should these test changes be in this branch? If I uncomment this I see `streamer_test.cxx` seems to be broken.

michele-brambilla: The main reason is that just a very small part of the Streamer can be tested in a unit test, since it requires Kafka to do almost anything. I'm still looking for a workaround to run without Kafka (that's why a few days ago I asked if you have some Kafka mock).
Since this PR introduces relevant changes  for the integration with NIKOS I'd like to defer the refactoring of the test in another PR.

matthew-d-jones: Ok. For Mantid I used mocks and fakes of the Kafka interface. Here are the fakes:
https://github.com/mantidproject/mantid/blob/master/Framework/LiveData/test/KafkaTesting.h
and some of the tests are here:
https://github.com/mantidproject/mantid/blob/master/Framework/LiveData/test/KafkaEventStreamDecoderTest.h
If you go with this approach it might make sense to do it after Mark's suggested refactoring of KafkaW into separate consumer and producer.

dominikwerder:
It seems to me that this feature branch was branched of the (volatile and convenience-only) `bleeding` branch and therefore shows a weird commit history. It would be great to rebase this feature branch on `master`.

src/Status.cxx
```diff
@@ -2,120 +2,127 @@
...
 const FileWriter::Status::MessageInfo &
-FileWriter::Status::MessageInfo::message(const double &message_size) {
-  std::lock_guard<std::mutex> lock(mutex_);
-  auto size = message_size * 1e-6;
-  atomic_add(Mbytes_, size);
-  atomic_add(Mbytes_square_, size * size);
-  atomic_add(messages_, 1.0);
-  atomic_add(messages_square_, 1.0);
+FileWriter::Status::MessageInfo::message(const double &MessageSize) {
+  std::lock_guard<std::mutex> Lock(Mutex);
```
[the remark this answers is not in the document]
michele-brambilla: In principle we can live without the lock, nevertheless in that case there could be inconsistencies. Since ``message`` (and ``error``) method and ``+=`` operator are used in different threads the data can be out of sync. Of course we can live with that, could happen that sometimes numbers will be a little silly.
I discussed this (and the next) comment with Dominik: he points out that in order to be completely consistent we need to use the same lock in ``operator+=`` and ``Report::produce_single_report``. 
Are you in favour of using the locks and trade a (possibly negligible, to be evaluated) slowdown with consistency in the statistic or no lock and possibility of inconsistencies? Of course these  inconsistencies are not critical at all, the report we produce is a just plus.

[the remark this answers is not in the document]
matthew-d-jones: The only operations the lock protects here are `atomic_add`, so are you saying `atomic_add` is not actually atomic?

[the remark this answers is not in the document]
michele-brambilla: The ``atomic_add`` has no inconsistencies, what I'm worried about is the lack of synchronization among the different variables which are used to compute the messages statistic. I'll try to explain myself with an example:
 
1. thread1 consumes a message and calls  ``MessageInfo::message``. At the same time thread 2 is sunning ``Report::produce_single_report``;
2. T1 executes ``atomic_add`` on ``Mbytes`` and ``MbytesSquare``. T2 executes ``StreamMasterInfo::add`` on the same data;
3. before T1 executes ``atomic_add`` on ``Messages`` T2 is executes  ``atomic_add`` on all the fields.
4. while T1 finishes all the  ``atomic_add`` T2 produces a message with some values at the current "iteration" and some at the previous.

Now, there is no inconsistency on the ``atomic_add`` itself but among the "iteration" at which the different values are considered.
Even worse: in the case above (without the locks) T2 could call ``MessageInfo::reset`` before T1 executes ``atomic_add`` on  ``Messages`` and ``MessagesSquare``.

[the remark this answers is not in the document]
matthew-d-jones: That's true, but all these operations are additions so it won't lead to a large cumulative error, right? If so, maybe it is better to prioritise performance of the statistics code over it being completely accurate? I don't know the requirements for the status information so I leave the decision to you.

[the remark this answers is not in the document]
dominikwerder: When we looked at the code, I suggested adding the required locks also on the reader side in order to calculate consistent values, because it's easier to remove locks later if it ever turns out to be a bottleneck, than to add locks in retrospect.
If our goal is only some form of eventual consistency, then like you said we would not need any locks.
The update frequency of the statistics is quite low, so my feeling is that we can afford the price for the consistency. If benchmarks tell otherwise, we can still lower our consistency standards.
So I would vote for keeping the locks at the moment.

michele-brambilla:
Thanks for your comments and suggestions. I made the changes you request except remove the locks. I also completely removed references to ``streamer_test`` which will be part of a different feature branch.

matthew-d-jones:
Shall we merge or do @dominikwerder or @nikhilbiyani want to take a final look?

michele-brambilla:
After a discussion with Nikhil and Dominik some changes are requested

michele-brambilla:
I did some changes in the implementation of the stop mechanism. Now it can properly rewrite history and stop even if any stream doesn't send data. If you want to give a look I think that is ready to be merged.

zjttoefs:
This sounds good. Since this is a complicated matter, I would like to see some documentation (specification) what the code actually attempts to implement and what should be considered expected behaviour from the user/NICOS point of view.
Working that out from the code is too error prone, at least in my head.

nikhilbiyani:
From NICOS point of view, following use cases are seen until now:

- Time/count preset: Start the file writing when the counting started and stop when it finishes. NICOS now issues two separate commands one to start the file writing and another one to stop it - both of them have their respective time stamps in the command.

- Rewrite data from past (Rewrite History): Start and stop time when the counting was on are stored in NICOS cache along with device values when the counting began. Now, when the user, for any reason, wants to rewrite the history, NICOS pulls up this list (list of start and stop times when the counting was on) from cache and issues one command for each element of the list. This command contains both start and stop time.

Each of these issued commands provides a nexus structure which may have a stream object.

src/MainOpt.cxx
```diff
@@ -49,16 +49,19 @@ int MainOpt::parse_config_json(std::string json) {
   if (auto o = get_int(&d, "status-master-interval")) {
     status_master_interval = o.v;
   }
-  if (auto o = get_object(d, "kafka")) {
```
dominikwerder: Where the options below the `kafka` key previously used for anything else than `stream-master` ?

src/CommandHandler.cxx
```diff
@@ -44,12 +44,12 @@ std::string find_broker(rapidjson::Document const &d) {
   return std::string{"localhost:9092"};
 }
 
-ESSTimeStamp find_time(rapidjson::Document const &d, const std::string &key) {
+KafkaTimeStamp find_time(rapidjson::Document const &d, const std::string &key) {
```
dominikwerder: Can we now completely remove the ESSTimeStamp typedef?

src/MainOpt.cxx
michele-brambilla: I don't remember was used for anything else; going back in the commit history nothing seems to be using it

michele-brambilla:
As Dominik suggested I removed all the types defined as aliases of ``std::millisecond`` and ``std::nanosecond`` (the latter is never used). This in turn let me remove ``utils.h`` which was mostly used for these definitions.

dominikwerder:
Builds and passes for me.

PR #98, opened 2018-01-16
68 configurable file output path

### LINKED ISSUE #68: Configurable file output path
I'd like to be able to provide an output path for the produced NeXus files via the configuration file and command line options.
This would allow me to mount a directory from the host machine as the output directory for the writer running in a docker container.

dominikwerder:
A prefix for the HDF output filename given in the JSON write command can now be specified either on the command line via `--hdf-output-prefix` or in the JSON configuration file under key `hdf-output-prefix`.
Added the parsing logic, unit tests and mentions the new option in README and the command line help.
Closes #68.

### DIFF
--- a/src/tests/HDFFile.cxx
+++ b/src/tests/HDFFile.cxx
@@ -23,6 +23,12 @@ using std::chrono::steady_clock;
 using std::chrono::milliseconds;
 using std::chrono::duration_cast;
 
+void merge_config_into_main_opt(MainOpt &main_opt, string jsontxt) {
+  rapidjson::Document cfg;
+  cfg.Parse(jsontxt.c_str());
+  main_opt.config_file = merge(cfg, main_opt.config_file);
+}
+
 // Verify
 TEST(HDFFile, create) {
   auto fname = "tmp-test.h5";
@@ -75,9 +81,7 @@ public:
     {
       std::string jsontxt =
           fmt::format(R""({{"hdf-output-prefix": "{}"}})"", hdf_output_prefix);
-      rapidjson::Document cfg;
-      cfg.Parse(jsontxt.c_str());
-      main_opt.config_file = merge(cfg, main_opt.config_file);
+      merge_config_into_main_opt(main_opt, jsontxt);
       main_opt.hdf_output_prefix = hdf_output_prefix;
     }
     rapidjson::Document json_command;
@@ -168,11 +172,7 @@ public:
 
   static void create_static_dataset() {
     MainOpt &main_opt = *g_main_opt.load();
-    {
-      rapidjson::Document cfg;
-      cfg.Parse(R""({})"");
-      main_opt.config_file = merge(cfg, main_opt.config_file);
-    }
+    merge_config_into_main_opt(main_opt, R""({})"");
     rapidjson::Document json_command;
     {
       using namespace rapidjson;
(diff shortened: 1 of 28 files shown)


src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
+
+    auto cmd = json_to_string(json_command);
+    auto fname = get_string(&json_command, "file_attributes.file_name");
+    ASSERT_GT(fname.v.size(), 8);
+
+    FileWriter::CommandHandler ch(main_opt, nullptr);
+    Msg msg;
+    msg.data = (char *)cmd.data();
+    msg.size = cmd.size();
+    ch.handle(msg);
```
matthew-d-jones: lines 149 to 152 could be cleaned up as
```
ch.handle({(char *)cmd.data(), cmd.size()});
```
Is there a reason `Msg.data ` shouldn't be `const char*` or even `Msg` be a `std::string`?

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
+    ASSERT_EQ(ch.file_writer_tasks.size(), (size_t)1);
+    {
+      string cmd("{\"recv_type\":\"FileWriter\", "
```
matthew-d-jones: `cmd` hides your previous `cmd` in upper scope. I think this code block should be extracted to a separate function, especially as it is repeated elsewhere.

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
 
+  static void create_static_file_with_hdf_output_prefix() {
+    using namespace FileWriter;
+    using std::array;
+    using std::vector;
+    using std::string;
+    MainOpt &main_opt = *g_main_opt.load();
+    std::string hdf_output_prefix = "tmp-relative-output";
+    {
+#ifdef _MSC_VER
```
matthew-d-jones: Thanks for thinking about running this on Windows, but I don't think it is worth adding untested bits of code like this. I'm going to talk with Freddie about whether we would need to run the writer on Windows at ISIS, because I think we should either not support it at all or have a Windows Jenkins build.

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
+      auto &j = json_command;
+      auto &a = j.GetAllocator();
+      j.SetObject();
+      Value nexus_structure;
+      nexus_structure.SetObject();
+
+      Value children;
+      children.SetArray();
+
+      {
```
matthew-d-jones: There are quite a few lines of json which need to be there but which aren't actually relevant to the test and make it less obvious to the reader what is being tested. Would it be possible to have a "core" json description in a function to which you add the specific bit you want to test in each test case?

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
+                 "\"cmd\":\"file_writer_tasks_clear_all\", "
+                 "\"job_id\":\"000000000dataset\" }");
+      ch.handle({(char *)cmd.data(), cmd.size()});
+    }
+    main_opt.hdf_output_prefix = "";
+
+    // Verification
+    auto fid = H5Fopen((hdf_output_prefix + "/" + fname.v).c_str(),
+                       H5F_ACC_RDONLY, H5P_DEFAULT);
+    ASSERT_GE(fid, 0);
```
matthew-d-jones: Isn't this `ASSERT` sufficient? The purpose of this test is to check the file is written to the requested location. You have other tests for checking the content is as expected.

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
     return event_index[i2] == cue_index;
   }
 
+  static void create_static_file_with_hdf_output_prefix() {
+    using namespace FileWriter;
+    using std::array;
+    using std::vector;
+    using std::string;
```
matthew-d-jones: These three `using std` statements can be removed; you aren't using them.

src/tests/HDFFile.cxx
```diff
@@ -66,6 +66,116 @@ class T_CommandHandler : public testing::Test {
...
+    {
+#ifdef _MSC_VER
+      // NOTE just a suggestion, not tested on windows so far, we have no
+      // windows here.
+      _mkdir(hdf_output_prefix.c_str());
+#else
+      mkdir(hdf_output_prefix.c_str(), 0777);
+#endif
+    }
+    {
```
matthew-d-jones: How about refactoring this code block to a function which takes the main_opt reference and a string to use as the config_file contents? This would deduplicate some code and declutter a couple of your test cases.

PR #99, opened 2018-01-18
Write attributes on file top level

dominikwerder:
Write standard top-level HDF attributes, like creation time/date, filename, HDF version, into the output HDF file.

### DIFF
--- a/src/tests/HDFFile.cxx
+++ b/src/tests/HDFFile.cxx
@@ -3,8 +3,8 @@
 #include "../HDFFile_impl.h"
 #include "../KafkaW.h"
 #include "../MainOpt.h"
-#include "../helper.h"
 #include "../h5.h"
+#include "../helper.h"
 #include "../schemas/ev42/ev42_synth.h"
 #include "../schemas/f142/f142_synth.h"
 #include <array>
@@ -544,7 +544,8 @@ public:
         }
       }
       auto t2 = CLK::now();
-      LOG(Sev::Debug, "processing done in {} ms", duration_cast<MS>(t2 - t1).count());
+      LOG(Sev::Debug, "processing done in {} ms",
+          duration_cast<MS>(t2 - t1).count());
       LOG(Sev::Debug, "finishing...");
       {
         string cmd("{\"recv_type\":\"FileWriter\", "
@@ -552,8 +553,10 @@ public:
         ch.handle({(char *)cmd.data(), cmd.size()});
       }
       auto t3 = CLK::now();
-      LOG(Sev::Debug, "finishing done in {} ms", duration_cast<MS>(t3 - t2).count());
-      LOG(Sev::Debug, "done in total {} ms", duration_cast<MS>(t3 - t1).count());
+      LOG(Sev::Debug, "finishing done in {} ms",
+          duration_cast<MS>(t3 - t2).count());
+      LOG(Sev::Debug, "done in total {} ms",
+          duration_cast<MS>(t3 - t1).count());
     }
 
     if (!do_verification) {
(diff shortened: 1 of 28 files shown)


matthew-d-jones:
I've removed reference to the `NeXus_version` attribute from the PR description because it is not implemented in the changes. According to the NeXus documentation it should "Only used when the NAPI has written the file" anyway.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
     H5Fclose(fid);
   }
 
+  static void write_attributes_at_top_level_of_the_file() {
+    using namespace FileWriter;
+    using std::array;
+    using std::vector;
+    using std::string;
```
matthew-d-jones: These `using std`s can be removed.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+    MainOpt &main_opt = *g_main_opt.load();
+    {
```
matthew-d-jones: Same comment as in #98, this block can be a reusable function

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+    })"");
+    auto cmd = json_to_string(json_command);
+    auto fname = get_string(&json_command, "file_attributes.file_name");
+    ASSERT_GT(fname.v.size(), 8);
+
+    FileWriter::CommandHandler ch(main_opt, nullptr);
+    Msg msg;
+    msg.data = (char *)cmd.data();
+    msg.size = cmd.size();
+    ch.handle(msg);
```
matthew-d-jones: Could use `ch.handle({(char *)cmd.data(), cmd.size()})` instead.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+    ASSERT_EQ(ch.file_writer_tasks.size(), (size_t)1);
+    {
```
matthew-d-jones: Can use the same `job_id` as in the other tests, then this is also a repeated code block with can be refactored to a separate function.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+        "cmd": "file_writer_tasks_clear_all",
+        "job_id": "832yhtwgfskdf"
+      })"");
+      ch.handle({(char *)cmd.data(), cmd.size()});
+    }
+
+    // Verification
+    auto fid = H5Fopen(string(fname).c_str(), H5F_ACC_RDONLY, H5P_DEFAULT);
+    ASSERT_GE(fid, 0);
+    {
```
matthew-d-jones: I would also suggest putting these in separate functions to improve readability, they could also be useful in other tests.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+      auto dtype = H5Aget_type(attr);
+      ASSERT_EQ(H5Tget_class(dtype), H5T_STRING);
+      std::vector<char> buf;
+      buf.resize(128);
+      H5Aread(attr, dtype, buf.data());
+      buf[buf.size() - 1] = 0;
+      ASSERT_EQ(string("Hello Attribute"), buf.data());
+      H5Aclose(attr);
+    }
+    {
```
matthew-d-jones: If this was a function it could also be used to check the `file_name` attribute.

src/tests/HDFFile.cxx
```diff
@@ -286,6 +286,91 @@ class T_CommandHandler : public testing::Test {
...
+      ASSERT_EQ(H5Tget_class(dtype), H5T_STRING);
+      auto dtype_mem = H5Tget_native_type(dtype, H5T_DIR_ASCEND);
+      std::vector<char> buf;
+      buf.resize(128);
+      H5Aread(attr, dtype_mem, buf.data());
+      buf[buf.size() - 1] = 0;
+      ASSERT_EQ(h5_version_string_linked(), buf.data());
+      H5Aclose(attr);
+    }
+    H5Fclose(fid);
```
matthew-d-jones: Should `file_time` always be present? If so, you could test that there is an attribute with that name present.

src/tests/HDFFile.cxx
dominikwerder: True, there is quite some opportunity for factoring here, but I keep that for later.

PR #102, opened 2018-01-18
Conan improvements

martukas:
* updated docker images
* updated conan packages
* added streaming-data-types conan package
* changed CMake and source files to use generated schema headers from conan rather than by relative path
* changed readme to reflect the new reality
* added .gitignore for builds in subdirectories + some IDE files


cmake/FindStreamingDataTypes.cmake
```diff
@@ -1,53 +0,0 @@
...
-# Make your target depend on 'flatbuffers_generate'.
-# When you call this using find_package, you can specify a minimum version
-# for the streaming data types:
-# find_package(streaming_data_types COMPONENTS <minimum-git-commit>)
-# you must then make your target depend on 'check_streaming_data_types'
-# so that it can abort compilation if streaming data types does not suffice.
-# Usage of this package requires of course the git suite to be available.
-# Contact: Dominik Werder
-
-find_path(path_include_streaming_data_types NAMES schemas/f141_epics_nt.fbs HINTS
```
dominikwerder: When was it decided that we do not support non-Conan builds any longer?
Removing this CMake module prevents a non-Conan build, or not?

src/schemas/ev42/ev42_rw.cxx
```diff
@@ -4,7 +4,7 @@
 #include "../../HDFWriterModule.h"
 #include "../../h5.h"
 #include "../../helper.h"
-#include "schemas/ev42_events_generated.h"
+#include "ev42_events_generated.h"
```
dominikwerder: It would be preferable to keep the `schemas/` prefix here, can that be accomplished using a Conan-provided `streaming-data-types`?

michele-brambilla:
First of all, I like the idea of having ``streaming-data-types`` as a conan package.
Having said that, I have a concern. What happens if I make some local change to ``streaming-data-types`` and want to use it in ``kafka-to-nexus``? In my understanding conan will always check out  the ``3e85b23`` commit. I should either modify the ``conan-streaming-data-types``
repository (and this will affect everyone else) or hack the local _.conan/data/..._, am I correct?

I don't know if this is possible: can the value of ``version`` and ``self.run("git checkout 3e85b237...")`` in ``conan-streaming-data-types/conanfile.py`` be determined from the content of conanfile.txt? To give an example, if I want to use the master repository I could write
```
streaming-data-types/master@ess-dmsc/stable
```
or to use commit af02b8d
```
streaming-data-types/af02b8d@ess-dmsc/stable
```
In this case the above situation is not an issue, it would be enough to change ``kafka-to-nexus/conan/conanfile.txt``.

matthew-d-jones:
Given that the decision to use Conan has been made we shouldn't keep any additional complexity to support not using it, so I think this is ok.

@michele-brambilla When testing/developing new schema or library versions I think using the commit as the version is a good idea, but we should have a non-stable channel for those recipes. Perhaps "testing"?
```
streaming-data-types/af02b8d@ess-dmsc/testing
```

matthew-d-jones:
As there are many merge conflicts with master and we have different solutions in `forward-epics-to-kafka`, I'll make the changes in #130 rather than trying to sort out this branch.

PR #108, opened 2018-01-28
Issue 107

martukas:
What has changed:
* Using new docker image, most recent conan packages
* hdf5 is no longer directly included but through h5cpp, which is provided through conan. No additional CMake magic should be needed. Information on HDF5 capabilities such as MPI and SWMR should be provided through h5cpp. If you need additional flexibility of features, please forward them to us so we can include them in the next release of h5cpp
* HDFFile and related tests have been refactored to use h5cpp. Very few vestiges of direct H5 C API remain. HDFFile has decreased in size by roughly 30%. A lot more opportunities for refactoring come into view.
* some related functions have had their signatures changed, for example `FileWriterTask::hdf_init`
* all tests pass, except `streamer_test`, which I am not quite sure how to run correctly
* exceptions are used in many places where return values were previously used. The new C++11 nested_exception is particularly good for providing a traceback stack. HDF5 error stack is thrown in the same way. You are welcome to use the convenience function `std::string print_nested(const std::exception&)` to unwrap such an exception for printing to logger or whatnot.
* explicit closing of H5 handles no longer necessary as h5cpp handles all the resource allocation. Many asserts are also not needed, because in the course of accessing HDF5 resources through h5cpp, many such checks are already in place and will throw exceptions in case of failure.

Outstanding questions:
* What is the role of `groups` in this function?:
```
void HDFFile::init(rapidjson::Value const &nexus_structure,
                  std::vector<StreamHDFInfo> &stream_hdf_info,
                  std::vector<hdf5::node::Group> &groups)
```
As far as I can tell, groups are only passed in to have their handles released (with some complaining about it). Given that all instances of `hid_t` are now handled by h5cpp, this may not be necessary? Do we have some other plans for this parameter?
* In `HDFFile::create_hdf_structure` there are three `Group` instances (`hdf_this`, `hdf_next_parent`, `gid`) defined that all eventually point to the same object, with no logic that ever differentiates them in any way. Are they at all needed there or could one variable suffice?


src/CMakeLists.txt
```diff
@@ -33,23 +32,21 @@ endif()
 
 set(compile_defs_common "")
 
-
 set(path_include_common
+	${PROJECT_SOURCE_DIR}/src
+	${PROJECT_BINARY_DIR}/src
```
matthew-d-jones: Indentation is a bit wonky here, two spaces rather than tabs please.

src/CMakeLists.txt
```diff
@@ -33,23 +32,21 @@ endif()
...
-  ${PROJECT_SOURCE_DIR}/src
-  ${PROJECT_BINARY_DIR}/src
 )
 
 set(libraries_common
   ${RDKAFKA_LIBRARIES}
-  ${HDF5_C_LIBRARIES}
   ${CURL_LIBRARIES}
   pthread
+	h5cpp
```
matthew-d-jones: tab->2 spaces

src/FileWriterTask.cxx
```diff
@@ -55,26 +55,42 @@ void FileWriterTask::add_source(Source &&source) {
...
+    return -1;
   }
+
   return 0;
 }
 
-int FileWriterTask::hdf_close() { return hdf_file.close(); }
+void FileWriterTask::hdf_close()
+{
+  // nest another exception?
```
matthew-d-jones: I think you can remove this comment; doesn't look like there is value in passing the exception up here.

dominikwerder:
Regarding the `groups` parameter: You're right, this is a remainder from tracking down open HDF objects. Turned out to be not the group anyway, but yes, this can go away.
Should I push a commit for this, or would you like to edit that, or open a new issue?

dominikwerder:
Regarding `HDFFile::create_hdf_structure` Yes, the different variables containing the same HDF object id can be streamlined. Also here: Should I push a commit?

src/HDFFile.h
```diff
@@ -26,31 +27,93 @@ class WriteResult {
...
+  static std::vector<std::string> populate_fixed_strings(rapidjson::Value const *vals,
+                                                         size_t fix_at,
+                                                         hssize_t goal_size);
+
+  static void write_ds_string(hdf5::node::Group& parent, std::string name,
+                              hdf5::property::DatasetCreationList& dcpl,
+                              hdf5::dataspace::Dataspace& dataspace,
+                              rapidjson::Value const *vals);
+
+  static void write_ds_string_fixed_size(hdf5::node::Group& parent, std::string name,
```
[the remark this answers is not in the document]
martukas: I won't do anything about this for now. I kept the naming convention I found. Let's wait for the json refactor to make this pretty.

martukas:
I have addressed @matthew-d-jones review and in the process I also made the changes in response to @dominikwerder comments: the `groups` parameter is gone, as are duplicated references to generated node. I also resolved reported merge conflicts.

dominikwerder:
On a clean terminal and clean directory, when I do

```
conan install --build=outdated ../kafka-to-nexus/conan
cmake ../kafka-to-nexus
make
```

I still get

```
[ 84%] Linking CXX executable ../kafka-to-nexus
/usr/bin/ld: cannot find -lh5cpp
collect2: error: ld returned 1 exit status
```

martukas:
@dominikwerder looks like you are not activating the conan-generated virtual environment. Of course, the only thing that we can objectively agree on is that it succeeds building on Jenkins. If you can configure your system to be as close as possible to the docker system, it should work. Maybe @amues can help you figure out whatever nuances of the conan configuration on your particular system that are preventing it from building.

dominikwerder:
@martukas If you refer to `activate_*.sh` scripts, then they do not make a difference for me.

Could it be that the `Findh5cpp` is not robust enough? If Jenkins has `h5cpp` installed system-wide, then I see how that error disappears. But that is against the purpose of `conan`. We should link against that specific version in the conan cache, and not against a system-wide installation.

I made an effort in PR 113 to establish compatibility, what prevents us from using that?

> Of course, the only thing that we can objectively agree on is that it succeeds building on Jenkins.

It is very objective, too to say that it does not build on my desktop machine at PSI.

martukas:
Perhaps "objective" is not the right word. What I meant is not universally accessible for testing to all developers working on the code.

It is absolutely true that neither h5cpp nor it's Conan recipe are robust enough to support all systems. They support quite a few systems, however, if you look in those projects, no less than two different centos7 configurations, Fedora 25, plain Debian 9, Ubuntu 16.04 LTS, the latest non-LTS Ubuntu 17.10, macOS and (for now in a separate job, not in the same Jenkins pipeline) Windows. For all the software that we intend to build and run on multiple platforms, we perform such tests. Daquiri and EFU(pending merge) are also now tested with parallel builds. I believe this infrastructure is in place precisely to ensure what I called "objectivity". I will repeat the offer to add a docker image for an additional distribution. You can even get it going yourself by example of the jobs that generate docker images for the other distros. It's all in the repos. But we should agree that it is something we must explicitly support. We were under the impression that PSI had transitioned to Centos7. Is that not so?

Anyways, getting the full stack support for another distro will take some time, even if we decide to do so. In the mean time, what could we do to make this merge happen? Just from the reported linker error I cannot figure out what needs to be changed nor confirm that my fix works for for your system. I have merged your proposed changes from master, and reintroduced the `find_h5simple` cmake script. That it builds on Jenkins is all I can myself confirm. So, if you can fix it in some other way but also maintain that it builds on Jenkins, I will be satisfied.

dominikwerder:
I merged my build-system related changes from PR #113 and it builds now on Jenkins as well as on my desktop. Should I press merge?

martukas:
:+1:

dominikwerder:
OK, I'll fix the conflict and merge then..

PR #113, opened 2018-01-31
Pull in h5cpp as a dependency, supports Conan as well as non-Conan

dominikwerder:
In preparation for the switch to h5cpp, this commit declares for cmake h5cpp
and the required dependencies.
h5cpp depends on Boost, therefore this is also a dependency now.

Using the new minimal Findh5cpp.cmake we can support Conan-builds as well as
non-Conan-builds.

### DIFF
--- a/cmake/Findh5cpp.cmake
+++ b/cmake/Findh5cpp.cmake
@@ -1,0 +1,7 @@
+find_path(H5CPP_INCLUDE_DIR NAMES h5cpp/hdf5.hpp)
+find_library(H5CPP_LIBRARIES NAMES h5cpp)
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(H5CPP DEFAULT_MSG
+  H5CPP_INCLUDE_DIR
+  H5CPP_LIBRARIES
+)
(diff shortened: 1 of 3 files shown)


cmake/Findh5cpp.cmake
```diff
@@ -0,0 +1,7 @@
+find_path(H5CPP_INCLUDE_DIR NAMES h5cpp/hdf5.hpp)
```
martukas: Just curious as to why this file is needed? I thought that the `h5cpp-config.cmake`, which is added to `/usr/local/lib/cmake` should be capable of finding it if you installed h5cpp to system.
As I understand, most of the changes below are for the sake of the older centos sytems?

dominikwerder: The rationale was to be explicit about finding the headers, also in the case when `h5cpp` is not installed in a system location.

> As I understand, most of the changes below are for the sake of the older centos sytems?

If you refer to adding the header directory to `path_include_common` and the libraries to `libraries_common` then this is to support both Conan as well as non-Conan builds in a, at least in my subjective and certainly biased opinion, minimal and non-magic way.

Issue #116, opened 2018-02-05
Do not overwrite existing HDF files


dominikwerder:
Do not overwrite existing HDF files.

https://jira.esss.lu.se/browse/DM-712

Issue #87, opened 2017-12-15
Error checking in HDFFile.cxx


mkoennecke:
There is very little error checking in HDFFile.cxx. This is bad! HDF5 calls can and will go wrong. Though most problems occur when trying to write, even when something else is wrong. Anyway, this needs to improve. The aim is to log the problem and continue if at all possible.

PR #117, opened 2018-02-05
Do not overwrite existing HDF file, but produce error instead

dominikwerder:
Do not overwrite existing HDF file, but produce error instead.

Adjust test cases to `unlink` an existing file before running the test.

Updates #116

### DIFF
--- a/src/HDFFile.cxx
+++ b/src/HDFFile.cxx
@@ -835,7 +835,7 @@ int HDFFile::init(std::string filename, rapidjson::Value const &nexus_structure,
   auto fcpl = H5Pcreate(H5P_FILE_CREATE);
   auto fapl = H5Pcreate(H5P_FILE_ACCESS);
   set_common_props(fcpl, fapl);
-  hid_t h5file = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, fcpl, fapl);
+  hid_t h5file = H5Fcreate(filename.c_str(), H5F_ACC_EXCL, fcpl, fapl);
   if (h5file < 0) {
     std::array<char, 256> cwd;
     getcwd(cwd.data(), cwd.size());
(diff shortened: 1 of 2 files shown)


SkyToGround:
Although this behaviour is (possibly) a slight improvement over the previous implementation, this is NOT the way to go. Throwing an error if the file already exist is a somewhat hostile behaviour. The application should try to recover if possible. The better implementation is to check if the file exists and attempt to use another file name if it does. For example, the MacOSX Finder appends a number to the file name if the file already exist.

dominikwerder:
@SkyToGround Thank you for your opinion. Regarding the PR, we first stop the overwrite. Whether or not we should try again with an automatic filename change should be configurable and will be addressed in another PR.

SkyToGround:
@dominikwerder I can see no reasonable argument (other than writing less code) for not having the default behaviour be that the application tries to recover by using an alternative file name. Care to elaborate why you think throwing an error is better?

zjttoefs:
Depends on who is in charge of the file name. As I understand it the filename is given to the writer. If that is not an acceptable choice, throwing an error is not unreasonable. The file could not be writable for other reasons (file system full or read-only) so dealing with errors is required.
If it was for the file writer to pick a working name in the first place, then I would agree with @SkyToGround.

SkyToGround:
@zjttoefs I agree that if it is not possible to create a file in the specified directory then an error should be thrown. Although it is not up to the file writer to pick the file name, mistakes will be made and having the file writer attempt to recover from those mistakes as gracefully as possible and pick another file name is the better option. This way, the probability of data loss is minimised. See also #87.

Seeing as the safer option requires relatively little extra code (depending on how you decide to implement it), there is not much downside to writing the "correct" implementation.

PR #121, opened 2018-02-09
Added coverage recipe

mattclarke:
Added coverage support to the build.


matthew-d-jones:
Should work after update to gocvr version in centos-gcc docker image: https://github.com/ess-dmsc/docker-centos7-gcc6-build-node/pull/1

PR #125, opened 2018-02-23
Issue 124

martukas:
* Pretty much each step is factored out into a function
* Stages only for each platform, since you can only explore the parallel build results in Blue Ocean anyway
* More informative exceptions
* On failure, shoots a message off to slack
* Does an actual build on Fedora
* Added Ubuntu17 and macOS builds
* When files are copied out of containers, it's into dedicated dirs for those images, to avoid clashes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /cmake-build-*/
 .idea/
 *.user
+pipeline.gdsl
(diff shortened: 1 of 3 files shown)


matthew-d-jones:
Made #127 with cleaner history

PR #127, opened 2018-02-26
Issue 124 

### LINKED ISSUE #124: Refactor Jenkins pipeline script
The Jenkins script has a few problems:
* Named stages unnecessary for Blue Ocean and actually produce buggy info in traditional view when names collide between parallel scripts.
* Downloading of schema is no longer necessary with conan.
* Some steps can be factored out into functions, per example of other dmsc projects

matthew-d-jones:
closes #124

>@martukas :
> -    Pretty much each step is factored out into a function
>-    Stages only for each platform, since you can only explore the parallel build results in Blue Ocean anyway
>-    More informative exceptions
>-    On failure, shoots a message off to slack
>-    Does an actual build on Fedora
>-    Added Ubuntu17 and macOS builds
>-    When files are copied out of containers, it's into dedicated dirs for those images, to avoid clashes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /cmake-build-*/
 .idea/
 *.user
+pipeline.gdsl
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,18 +3,18 @@ clangformat_os = "fedora"
 test_and_coverage_os = "centos7-gcc6"
 
 images = [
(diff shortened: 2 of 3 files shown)


matthew-d-jones:
Thanks for the changes @martukas 
Addition of Mac build to CI is good to support development on Macs at ESS. Refactoring and improved error handling looks good.

PR #129, opened 2018-03-01
Optional start stop time

dominikwerder:
During command parsing, make the start and stop time optional again.
Regression got introduced by PR #123.


matthew-d-jones:
@dominikwerder Can this be covered by tests to prevent regression in the future?

PR #135, opened 2018-03-09
Add conan CMake options

### LINKED ISSUE #134: Add conan cmake options
Add CMake variables to allow using an externally generated conan cmake dependencies file, or to use a profile other than `default`. In readme, document requirement for conan to be in the path if the automatic option is used.

matthew-d-jones:
Closes #134

Creates the following options for using conan:
- By default conan is used to download and configure dependencies, this is done automatically by CMake.
conan is required to be installed and in the `path`.
- conan can be run manually to generate a `conanbuildinfo.cmake` file in the build directory, to use this specify the CMake parameter `CONANBUILDINFO_ENABLE=true`.
- conan can be disabled by specifying the CMake parameter `CONAN_DISABLE=true`. CMake will try to find system installed libraries or paths can be specified manually.

These options are documented in the README.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,20 +11,23 @@ if(CMAKE_COMPILER_IS_GNUCXX AND COV)
 endif()
 
 SET(CONAN_PROFILE "default" CACHE STRING "Name of conan profile to use, uses default by default")
-if(CONANBUILDINFO_ENABLE)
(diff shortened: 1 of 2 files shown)


matthew-d-jones:
@michele-brambilla You're right, that sounds better, try this.

michele-brambilla:
Perfect, can we merge?

PR #136, opened 2018-03-09
Fix long integer bug

### LINKED ISSUE #133: Long integers cause error
For example:
```json
{
    "name": "features",
    "values": 10138143369737381149,
    "type": "dataset",
    "dataset": {
        "type": "uint64"
    }
}
```

matthew-d-jones:
Fixes #133 

Fixes a bug with getting long integers from the json command. This stopped `features` datasets being written for the example ESS NeXus files.

Added test for writing each of the four integer types.

### DIFF
--- a/src/tests/HDFFile.cxx
+++ b/src/tests/HDFFile.cxx
@@ -1262,14 +1262,14 @@ TEST_F(T_CommandHandler, createStaticDatasetWithLongIntegerValue) {
 
   std::string CommandWithLongInt = R""({
       "children": [
-      {
-        "name": "features",
-        "values": 10138143369737381149,
-        "type": "dataset",
-        "dataset": {
-          "type": "uint64"
-      }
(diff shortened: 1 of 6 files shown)


src/tests/HDFFile.cxx
```diff
@@ -1265,3 +1255,43 @@ TEST_F(T_CommandHandler, data_f142) { T_CommandHandler::data_f142(); }
 TEST_F(T_CommandHandler, dataset_static_1d_string_variable) {
   T_CommandHandler::dataset_static_1d_string_variable();
 }
+
+template <typename T>
+void testType(FileWriter::HDFFile &TestFile,
```
dominikwerder: Would a function name like `verifyWrittenDatatype` be more appropriate? Because, all datatypes are are written to the same HDF file in the same test run.

PR #143, opened 2018-03-12
Streamer uses KafkaW

michele-brambilla:
This PR introduces two main changes

1. the ``streamer``now  uses ``KafkaW`` instead of librdkafka; 
2. the ``connect()`` task is now executed in task-based parallelism instead of thread-based;

and a new feature

3. added tests for the ``Streamer``.

Consequences:

1) makes the filewriter implementation more uniform. ``KafkaW`` is the only explicit connection with librdkafka;
2) let us remove explicit locks and condition variables. The flow is now more linear;
3) needs the fake librdkafka implementation to be specified using ``-DFAKE_RDKAFKA=<path>`` when invoking ``cmake``.


michele-brambilla:
Please note that some cleanup is still required. Nevertheless I would like to have your opinion and any suggestion.

src/Msg.h
```diff
@@ -77,7 +77,7 @@ class Msg {
     return msg;
   }
 
-  static Msg fromKafkaW(std::unique_ptr<KafkaW::Msg> KafkaWMsg) {
+  static Msg fromKafkaW(std::unique_ptr<KafkaW::Msg> &&KafkaWMsg) {
```
matthew-d-jones: It looks to me like `fromKafkaW` should take ownership of the the `KafkaWMsg` pointer, and therefore passing it by value makes sense to me, why the change?

src/Streamer.cxx
```diff
@@ -206,28 +218,31 @@ void FileWriter::Streamer::connect(const std::string &TopicName) {
...
-    return;
-  }
-  LOG(Sev::Debug, "createTopicPartition");
-  RunStatus = assignTopicPartition();
-  if (RunStatus != SEC::no_error) {
-    return;
-  }
+  ConsumerW.reset(new KafkaW::Consumer(Settings));
+  ConsumerW->addTopic(TopicName);
+  ////
```
matthew-d-jones: Don't forget to clean up this

src/Msg.h
michele-brambilla: You're right, that should not have changed

michele-brambilla:
I have no more commit to push, any review and suggestion is welcome. I modified the initial message to give a better overview of the changes.

src/Streamer.cxx
```diff
@@ -1,321 +1,143 @@
...
-    return ProcessMessageResult::OK();
+  // wait for connect() to finish
+  try {
+    if (IsConnected.valid()) {
+      if (IsConnected.wait_for(std::chrono::milliseconds(1000)) !=
+          std::future_status::ready) {
+        LOG(Sev::Critical, "... still not ready");
+        return ProcessMessageResult::OK();
+      }
+      std::call_once(ConnectionStatus,
```
dominikwerder: Why is `std::call_once` necessary here?

src/tests/CMakeLists.txt
```diff
@@ -15,18 +16,33 @@ status_test.cxx
 status_writer.cxx
 command_handler.cxx
 stream_master.cxx
-MainOpt_tests.cxx
```
dominikwerder: Why are they removed?

src/Streamer.cxx
michele-brambilla: Not anymore, before the fture::get was outside of the if condition. Fixed.

src/tests/CMakeLists.txt
michele-brambilla: No reason, restored.

src/StreamMaster.h
```diff
@@ -45,9 +45,13 @@ template <typename Streamer> class StreamMaster {
...
-                        std::forward_as_tuple(d.topic()),
-                        std::forward_as_tuple(broker, d.topic(), options));
+      try {
+        Streamers.emplace(std::piecewise_construct,
+                          std::forward_as_tuple(d.topic()),
+                          std::forward_as_tuple(broker, d.topic(), options));
+      } catch (std::exception &e) {
+        LOG(Sev::Critical, "{}", e.what());
+      }
       Streamers[d.topic()].setSources(d.sources());
```
dominikwerder: Does this behave correctly if the code in `try` throws?

src/Streamer.cxx
```diff
@@ -1,321 +1,142 @@
...
+    }
+  } catch (std::exception &Error) {
+    LOG(Sev::Critical, "{}", Error.what());
   }
-  std::lock_guard<std::mutex> lock(
-      ConnectionReady); // make sure that connect is completed
 
+  // make sure that the connection is ok
+  // attention: connect() handles exceptions
   if (int(RunStatus) < 0) {
```
dominikwerder: Instead of C style cast, is it possible to have some `RunStatus.connectionOk()` ?

src/tests/streamer_options.cxx
```diff
@@ -57,52 +62,38 @@ void StreamerOptionsTest::CompareSteamerOptionsWith(
...
-  rapidjson::Document optj;
-  optj.Parse(R""({
-        "streamer" : {
-            }
-      })"");
-  ConfigureStreamerWithJson(optj);
-  FileWriter::StreamerOptions Defaults;
-  CompareSteamerOptionsWith(Defaults);
-}
+// TEST_F(StreamerOptionsTest, streamer_empty_json_doesn_t_change_defaults) {
```
dominikwerder: Can this commented test be removed?

src/CommandHandler.cxx
```diff
@@ -353,6 +351,8 @@ void CommandHandler::handleStreamMasterStop(std::string const &Command) {
   if (!MasterPtr) {
     return;
   }
+  LOG(Sev::Critical, "{}", Command);
```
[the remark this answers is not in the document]
michele-brambilla: Is there some other place where we can log (maybe in Debug mode only) the command? It could be helpful for debugging. If not, I would suggest to keep this log and set it to Sev::Debug.

src/Streamer.cxx
michele-brambilla: The two ``enums`` (StreamerErrorCode and StreamMasterErrorCode) are replaced with two classes. Now it is possible.

src/tests/streamer_options.cxx
michele-brambilla: :+1:

src/StreamMaster.h
michele-brambilla: The behaviour was in any case correct, nevertheless it is more clean as we discuss this morning.

PR #144, opened 2018-03-14
H5cpp in dataset writer

dominikwerder:
Remove usage of plain HDF C API in the file writer dataset writer.
A few remaining places are adressed in another branch.

### DIFF
--- a/src/h5.cxx
+++ b/src/h5.cxx
@@ -337,7 +337,7 @@ h5d_chunked_1d<T>::h5d_chunked_1d(std::string name, h5d ds_)
 
 template <typename T>
 h5d_chunked_1d<T>::h5d_chunked_1d(h5d_chunked_1d &&x)
-    : ds(move(x.ds)), dsp_wr(move(x.dsp_wr)) {}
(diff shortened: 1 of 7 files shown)


src/h5.cxx
```diff
@@ -4,242 +4,102 @@
...
-}
-
 void h5d::init_basics() {
-  herr_t err = 0;
-  type = H5Dget_type(id);
-  dsp_tgt = H5Dget_space(id);
-  ndims = H5Sget_simple_extent_ndims(dsp_tgt);
+  Type = Dataset.datatype();
+  DSPTgt = Dataset.dataspace();
+  ndims = DSPTgt.rank();
```
matthew-d-jones: Avoid implicit cast (`int` to `hsize_t`)

src/h5.h
```diff
@@ -15,46 +17,12 @@ using std::array;
 using std::vector;
 using std::string;
```
matthew-d-jones: I realise it is not in these changes, but now is a good time to fix it. Could these `using` statements go in the cxx file to avoid polluting the h5 namespace? Only `unique_ptr` and `string` are used in the header and not in many places.

PR #149, opened 2018-03-14
Extracts out ev42 schema from cxx into header file

### LINKED ISSUE #137: ev42_rw.cxx extract relevant part into .h file
For consistency with other classes.

rerpha:
fixes #137

Extracts out ev42 schema from cxx into header file, also removing using declarator for several types


rerpha:
needs header includes tidying up

src/schemas/ev42/ev42_rw.h
```diff
@@ -0,0 +1,61 @@
+#include "../../FlatbufferReader.h"
+#include "../../HDFWriterModule.h"
+#include "../../h5.h"
+
+namespace FileWriter {
+namespace Schemas {
+namespace ev42 {
+template <typename T> using uptr = std::unique_ptr<T>;
+
+struct append_ret {
```
matthew-d-jones: This `struct` is only used in the implementation of these classes so it doesn't need to be in the header.

PR #151, opened 2018-03-19
Use CLI11 for command line arguments

### LINKED ISSUE #114: Consider using CLI11 in place of getopt
The syntax of `CLI11` looks much nicer than `getopt`. It also has the advantage of being cross-platform; `getopt` is POSIX-only.
The library is header-only and bincrafters have released a conan package.

https://github.com/CLIUtils/CLI11
https://bintray.com/bincrafters/public-conan/cli11%3Abincrafters

matthew-d-jones:
Closes #114 

Please check I have not missed any options, some were undocumented in the README and `--help` text.
For now we continue to use the json configuration file, but we could switch to using the CLI11 ini file in the future (#152). This would simplify the code to a single definition of the options and default values.

I've also used CLI11 for options in `send-command` so please check that too.

There is no description for `teamid` in `--help`, this is not a regression and I've created an issue for it #160.


matthew-d-jones:
Investigating coverage decrease

matthew-d-jones:
@dominikwerder @michele-brambilla @SkyToGround This PR is now ready for review

conan/conanfile.txt
```diff
@@ -9,6 +9,7 @@ jsonformoderncpp/3.1.0@vthiery/stable
 graylog-logger/1.0.3-dm1@ess-dmsc/stable
 h5cpp/0.0.5@ess-dmsc/stable
 streaming-data-types/3e85b23@ess-dmsc/stable
+cli11/1.3.0@bincrafters/stable
```
michele-brambilla: Can you add the remote for ``cli11`` in the README ?

src/CLIOptions.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void setCLIOptions(CLI::App &App, MainOpt &MainOptions) {
+  // disable ini config file
+  App.set_config();
+  // and add option for json config file instead
+  App.add_option("--config-file", MainOptions.config_filename,
+                 "Specify a json file to set config")
+      ->check(CLI::ExistingFile);
+
+  add_option(
```
michele-brambilla: Is there any difference between ``App.add_option("...)`` and ``add_option(App,...`` ? If not, can we use only one of the two (to keep the style consistent)?

src/CMakeLists.txt
```diff
@@ -108,7 +108,7 @@ set(kafka_to_nexus_SRC
   MainOpt.cxx
   ${FMT_SRC}
   ${CMAKE_CURRENT_BINARY_DIR}/git_commit_current.cxx
-)
+  CLIOptions.cpp)
```
michele-brambilla: Since we're using ``.cxx`` for all the source files, can we change the extension for this file as well?

src/tests/CMakeLists.txt
```diff
@@ -16,8 +16,8 @@ status_test.cxx
 status_writer.cxx
 command_handler.cxx
 stream_master.cxx
+MainOptTest.cpp
```
michele-brambilla: Since we're using ``.cxx`` for all the source files, can we change the extension for this file as well?

src/CLIOptions.cpp
matthew-d-jones: The latter is implemented by me to allow adding `URI` options. It is not really an overload as the former is a method of `App` but the suggestion in the CLI11 docs is to use `add_option` for your own types.

src/CMakeLists.txt
matthew-d-jones: `.cpp` and `.h` are to be used according to LLVM guidelines.

conan/conanfile.txt
matthew-d-jones: Jonas already did that on `master`, see https://github.com/ess-dmsc/kafka-to-nexus#conan

matthew-d-jones:
@michele-brambilla Thanks for the review, I've pushed changes.

src/tests/CMakeLists.txt
SkyToGround: The long term goal is to change all file names to follow the LLVM standard. In order to not make the pull requests to big, changing the names of only files that are otherwise touched is the way to go.

PR #159, opened 2018-03-23
Removed undefined behaviour and, as a bonus, removed one global.

### LINKED ISSUE #158: Undefined behaviour in kafka-to-nexus.cxx
The signal handler (`signal_handler()`) in _kafka-to-nexus.cxx_ implements undefined behaviour. A list (that is not complete) of functions that you are allowed to call can be found here: http://man7.org/linux/man-pages/man7/signal-safety.7.html

Note that calling any of the `printf` family of functions in a signal handler is undefined behaviour.

SkyToGround:
Fixes #158

Fixed undefined behaviour, see issue #158.


SkyToGround:
The build failure is due to me having a newer version of clang-format than the build node. I will fix this within a few days

PR #163, opened 2018-03-28
Splitting tests into unit and integration tests and general clean-up of the test code

### LINKED ISSUE #49: Separate unit tests and integration tests

### LINKED ISSUE #162: Clean-up of unit tests
This is a general issue related to #49 as well as moving the unit tests to follow the LLVM standard.

After examining the code in _src/tests_ more closely, it turns out that the amount of integration testing in that code was not as large as I expected. However, it still needs some amount of work. I am currently working on this issue and plan to have a pull request done by early next week.

SkyToGround:
Fixes #49 
Fixes #162 

This is what I have done:
1. Split the tests into unit tests and integration tests.
2. Fixed failing test.
3. LLVM:ified the file names and test names (changed suffixes to .cpp and made names camel case).
4. Removed some dead code.
5. Moved test (json) data into separate directory and made the tests find that data using absolute paths.
6. Fixed null pointer access bug in the integration test code.
7.  Other minor fixes and improvements to the code.

The new _IntegrationTests_ target builds and runs though it crashes. Someone with more knowledge about the integration test should take a look at it. I do not believe I have broken anything in it.


SkyToGround:
@matthew-d-jones Although there are still a few tests that exercises multiple classes, I would still call those unit tests. I did move the tests that required external executables to be running to another target (_IntegrationTests_). Many tested functions do some amount of system testing as they are using the logging system, but that is somewhat unavoidable I think.

There are a few more changes that I would like to do but I have not in order to minimise the code changes:

1. There are some tests that does not actually test anything (e.g. `HDFFile.Create`).
2. Some tests could be simplified a bit.
3. Move integration and unit testing code into separate directories.

If we do change the name of the test-target back to "tests", then I it is my opinion that we should as a minimum also move the integration testing code to a separate directory. Though I think we should do that anyways.

What we could (and maybe should) do is to move the _HDFFile.cpp_ code to the integration testing target as that is the worst offender in terms of not conforming to containing unit tests.

PR #164, opened 2018-03-28
Fixed undefined behaviour and removed dead code

### LINKED ISSUE #156: AllocStd.h does not provide any useful functionality and is used to implement undefined behaviour
The code in AllocStd.h currently does nothing useful that can not be replaced with a single line of code (`new std::uint8_t[SomeSize]`) and should probably be removed for this reason alone. However, I also noted that AllocStd.h allocates memory using `malloc()`, which is then deallocated using `delete`. This is undefined behaviour.

Is there a good reason not to remove AllocStd.h?

SkyToGround:
Fixes #156 

Things changed in this pull request:
1. Removed the use of malloc/delete (undefined behaviour).
2. Removed dead code caused by switching out Alloc.h/AllocStd.h for `new`.
3. Minor CMake fix and fixes required by the above mentioned changes.

Not having been involved in the design of `FileWriter::Msg`, this class seems fragile and overly complicated to me. There might be a good reason for this but I want to point out that my gut reaction is that this class really could do with some unit tests and the use of `unique_ptr`.


matthew-d-jones:
@SkyToGround Thanks, I'll take a look at these changes soon. I agree with your comment on `Msg`, I mentioned a similar sentiment on the `#kafka_to_nexus` on Slack a few days ago.

src/Streamer.h
```diff
@@ -18,6 +18,7 @@
 
 #include "KafkaW/KafkaW.h"
 
+#include <chrono>
```
matthew-d-jones: Was this intentional?

src/DemuxTopic.h
```diff
@@ -3,6 +3,7 @@
 #include "Source.h"
 #include "TimeDifferenceFromMessage.h"
 #include "json.h"
+#include <chrono>
```
matthew-d-jones: Was this intentional?

src/Msg.h
```diff
@@ -160,10 +149,10 @@ class Msg {
       delete var.kafkaw_msg;
       break;
     case MsgType::Owned:
-      delete var.owned;
+      delete[] var.owned;
```
matthew-d-jones: `var.owned` was just initialised as a raw pointer, so this isn't paired with `new[]`. I think it was correct before.

src/Streamer.h
SkyToGround: Yes, this was intentional. I have not looked at the "include paths" but when I removed Alloc.h, some of the build nodes complained about missing std::chronological. With that said, you do have a point in that those include-statements could (should) probably be in the cpp-file.

src/DemuxTopic.h
SkyToGround: Yes, this was intentional. I have not looked at the "include paths" but when I removed Alloc.h, some of the build nodes complained about missing std::chronological. With that said, you do have a point in that those include-statements could (should) probably be in the cpp-file.

src/Msg.h
SkyToGround: `var.owned` is being set on line 31, the actual line is:
```c++
msg.var.owned = new char[len];
```

dominikwerder:
The allocator-related code was used for performance comparisons before.

PR #175, opened 2018-04-13
Simplified (unified) interface for registering file writing modules

### LINKED ISSUE #174: Simplify and unify interface for registering file writing modules
Based on issue #173 and discussions with Dominik and Tobias.

SkyToGround:
Closes #174 

This is what has been done:
* Modified the writer registation code to take a template argument for registering a new writer module instead of a `std::function`.
* Changed key type for reader module from `std::array<char,4>` to `std::string`.
* Added some error checking.
* Mode names clearer and LLVM-ified some names.
* Added unit tests. Probably too many for such simple code but I will not remove them now.
* Modified file name suffixes to _*.cpp_.


src/schemas/f142/f142_rw.cxx
```diff
@@ -149,11 +149,7 @@ uint64_t FlatbufferReader::timestamp(Msg const &msg) const {
 }
 
 FlatbufferReaderRegistry::Registrar<FlatbufferReader>
-    g_registrar_FlatbufferReader(fbid_from_str("f142"));
-
-FileWriter::HDFWriterModule::ptr HDFWriterModule::create() {
-  return FileWriter::HDFWriterModule::ptr(new HDFWriterModule);
-}
+    RegisterEPICSDataReaderModule("f142");
```
dominikwerder: Would e.g. `RegisterFlatbufferReaderModule` a less confusing name? The casual reader could think that we might do something with the EPICS API here.

Should be `static`.

src/schemas/f142/f142_rw.cxx
```diff
@@ -596,8 +592,8 @@ int32_t HDFWriterModule::flush() { return 0; }
 
 int32_t HDFWriterModule::close() { return 0; }
 
-HDFWriterModuleRegistry::Registrar
-    g_registrar_HDFWriterModule("f142", HDFWriterModule::create);
+HDFWriterModuleRegistry::Registrar<HDFWriterModule>
+    RegisterEPICSDataWriterModule("f142");
```
dominikwerder: Should be `static`.

Naming, like in the above comment?

src/schemas/f142/f142_rw.cxx
SkyToGround: I agree that the name could probably be improved but I strongly disagree that there is a confusion with the EPICS API here as the given types makes it clear that this is not the case. I do not think that `RegisterFlatbufferReaderModule` is an improvement as every single reader (and writer) is a flatbuffer reader/writer. As for static, I will have that fixed in the next commit.

src/schemas/f142/f142_rw.cxx
SkyToGround: See previous comment.

src/tests/ReaderRegistration.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+class ReaderRegistrationTest : public ::testing::Test {
+public:
+  void SetUp() override {
+    std::map<std::string, ReaderPtr> &Readers =
+        FlatbufferReaderRegistry::getReaders();
+    Readers.clear();
+  };
+
+  void TearDown() override{
```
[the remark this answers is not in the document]
dominikwerder: Well, I meant the `override{` without space, but never mind, we agreed that `clang-format` has the authority.

src/schemas/f142/f142_rw.cxx
SkyToGround: Naming things is hard, right? How about giving up and simply having the line be:
```
static HDFWriterModuleRegistry::Registrar<HDFWriterModule> Register("f142");
```
Most of the information we are interested in is already in that line anyways.

dominikwerder: I like that.

SkyToGround: Due to name collision, the line is instead:
```
static HDFWriterModuleRegistry::Registrar<HDFWriterModule> RegisterWriter("f142");
```

dominikwerder:
Thank you for the changes, LGTM

PR #194, opened 2018-05-23
React only on commands addressed to service

dominikwerder:
### Description of work

- On incoming filewriter command, check if it contains a `service_id`.
- If not, we currently accept the command for compatibility. Behavior can be changed in the future in one line.
- If yes, we check if it matches the `service_id` of the running instance before we accept the command.
- The `service_id` can be given to the filewriter via CLI option or configuration file.

### Issue

Closes https://jira.esss.lu.se/browse/DM-726

### Acceptance Criteria

Filewriter should accept command only if `service_id` matches, or not given in command.
Should get an integration test.

---

---

### DIFF
--- a/src/CommandHandler.cpp
+++ b/src/CommandHandler.cpp
@@ -395,6 +395,19 @@ void CommandHandler::handle(std::string const &Command) {
     LOG(Sev::Error, "Can not parse json command: {}", Command);
     return;
   }
+
+  if (auto x = find<std::string>("service_id", Doc)) {
+    if (x.inner() != Config.service_id) {
+      return;
+    }
+  } else {
+    // Currently, we interpret commands which have no service_id.
+    // In the future, we may want to ignore all commands which are not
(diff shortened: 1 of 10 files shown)


src/CommandHandler.cpp
```diff
@@ -395,6 +395,19 @@ void CommandHandler::handle(std::string const &Command) {
     LOG(Sev::Error, "Can not parse json command: {}", Command);
     return;
   }
+
+  if (auto x = find<std::string>("service_id", Doc)) {
+    if (x.inner() != Config.service_id) {
+      return;
```
matthew-d-jones: I think we should log when this happens, it'll make the circumstance of using an invalid `service_id` easier to debug than if all the file writers just silently ignore the command.

dominikwerder: I make it `Debug` level though, because I assume that this may happen all the time under normal operations if writers share a topic.

PR #195, opened 2018-05-24
Optional SWMR support

dominikwerder:
### Description of work

HDF SWMR can be enabled by adding to the filewrite command:

```
{ "use_hdf_swmr": true }
```

### Issue

Closes https://jira.esss.lu.se/browse/DM-412
Closes https://jira.esss.lu.se/browse/DM-728

### Acceptance Criteria

Files should get written as normal.
SWMR reading could be added to the integration tests.

---

---


dominikwerder:
According to the HDF presentation at the last ICALEPCS, the performance impact can be rather large, but this can be investigated in more detail.

Regarding the POSIX filesystem requirement: HDF does not say that SWMR writing works fine on a non-POSIX filesystem. How I read it, there could be not only inconsistent data on the reader side, but (for some reason I can not yet see) also the writing would not be guaranteed to work.

matthew-d-jones:
If the intention is to always use SWMR at the ESS then it makes sense to make it the default so that our testing is done with it enabled etc.

I hope no one is using a non-POSIX compliant filesystem anyway. Everyone is using XFS, ZFS, NTFS, Lustre... right?

matthew-d-jones:
oh APFS on OS X, is that POSIX compliant?

SkyToGround:
I am very confident that APFS is POSIX compliant. Interesting fact: the old HFS+ file system used by MacOSX is (generally) not POSIX compliant, but that is due to the way lower and upper case letters are handled.

PR #197, opened 2018-05-28
Restore streamer.cpp in tests

michele-brambilla:
### Description of work

Restore the tests for the Streamer class. See [README->Running tests](https://github.com/ess-dmsc/kafka-to-nexus#running-tests).

### Issue

Not an issue, ticket [DM-399](https://jira.esss.lu.se/browse/DM-399)

### Unit Tests

Tests handling of the following failures (and successes):
- missing constructor arguments
- configuration creation
- missing topic
- connection to the broker
- consumer timeouts

### Other

Here we want to test the Streamer response to network calls. In order to do that we need to fake the rdkafka library: such an implementation can be found in [ess-dmsc/librdkafka-fake](https://github.com/ess-dmsc/librdkafka-fake). Of course I'm open to alternatives.

---

---

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -25,22 +25,7 @@ set(UnitTests_INC
 HDFFileTestHelper.h
 )
 
-add_executable(UnitTests EXCLUDE_FROM_ALL ${UnitTests_SRC} ${UnitTests_INC})
-
-
-add_dependencies(UnitTests flatbuffers_generate)
-target_compile_definitions(UnitTests PRIVATE ${compile_defs_common})
-target_include_directories(UnitTests PRIVATE ${path_include_common})
-target_link_libraries(UnitTests ${GTEST_BOTH_LIBRARIES} ${libraries_common})
-
-get_filename_component(TEST_DATA_PATH "data/" ABSOLUTE)
-target_compile_definitions(UnitTests PRIVATE TEST_DATA_PATH="${TEST_DATA_PATH}")
-
-enable_testing()
-add_test(AllTests UnitTests)
-
-#-------------- Integration tests --------------
-
+# replace librdkafka with fake in order to test the Streamer
 set(libraries_testing ${libraries_common})
 if(NOT ${FAKE_RDKAFKA} STREQUAL "")
   message(STATUS "Build with fake librdkafka: " ${FAKE_RDKAFKA})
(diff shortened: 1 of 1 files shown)


matthew-d-jones:
The third of Jonas's bullet points ("dependency injection") is almost always my preferred method for testing. It is very flexible as you can construct and pass in whatever mock or fake you like for a given unit test or test suite. It can also have a positive impact on the code quality; preventing too tight coupling of separate classes.

However, using an external library like this is a great deal better than not testing this code at all. Also, in the future we can refactor the the functionality in `librdkafka-fake` into fake consumer and producer classes for injection, if we agree that is preferred.

Would it be possible to bring in `librdkafka-fake` as a conan package?

SkyToGround:
@matthew-d-jones I agree that using an external library is better than not testing at all. On the other hand, because of all the steps required with this implementation, this is very close to not testing at all. If we go for the solution where we use a fake librdkafka-implementation, then I think CMake and Conan (and/or Jenkins) should do it automatically.

mattclarke:
From a brief glance, it is just a case of needing to mock out `KafkaW::Consumer` or am I missing something more subtle?

SkyToGround:
From looking at the code yesterday, I would say that is exactly what is needed @mattclarke .

michele-brambilla:
@mattclarke you are correct. When the tests were first added the Streamer was using rdkafka directly. Later we decided to use KafkaW. Now mocking KafkaW would do the job.
But the problem then shifts to KafkaW, we must to rely on the fact that it deals properly with all the possible "failures" of rdkafka.

That was the idea behind the replacement of librdkafka. We can reproduce all the possible successes and failures without the need of a broker up and running.

mattclarke:
That is the advantage of mocking KafkaW::Consumer: it allows you to test the logic without needing a real broker and avoids the need for extra CMake logic.

For example, you can create a mock that simulates not finding a topic (not real code):

```
class MockConsumerThatDoesNotFindTopic {

bool topicPresent(const std::string &Topic) {
    // Always return false to simulate topic not being found
    return false;
}
};
```
There is an example of a simple mock being used in testing in the forwarder https://github.com/ess-dmsc/forward-epics-to-kafka/blob/master/src/tests/streams_tests.cpp

It might be worth talking to @matthew-d-jones and @rerpha about how they did this.

SkyToGround:
@michele-brambilla I agree. Fake librdkafka looks great for testing of `KafkaW::Consumer` and `KafkaW::Producer`. `FileWriter::Streamer` does not use librdkafka directly however (but instead the abstraction that is `KafkaW::Consumer`) and thus when testing this class, it is `KafkaW::Consumer` that should be faked/mocked.

SkyToGround:
@michele-brambilla This pull request can now be closed as I have implemented proper unit tests of the 'FileWriter::Streamer' class using mocking.

PR #211, opened 2018-06-15
Remove MainOpt constructor and add init

amues:
The presence of the constructor causes the file writer not to write data
to file in the integration test (the command is received, as the NeXus
file is created with the correct structure). Just commenting out the
contents of the constructor makes no difference. As a temporary
solution, an init function was added to initialise the variable that was
being initialised in the constructor. If the cause of the problem is
discovered and fixed, the constructor can be reintroduced.

### Description of work

See commit text above.

### Issue

Closes JIRA DM-927

### Acceptance Criteria

The build succeeds, unit tests pass and the integration test passes.

### Unit Tests

_src/tests/HDFFile.cpp_ and _src/tests/MainOptTest.cpp_ have been modified to call the new `init` function after the `MainOpt` struct is created.

### Other

For the sake of elegance and resource management, the new `init` should be replaced by the constructor after the problem is understood and fixed.

---

---

### DIFF
--- a/src/tests/HDFFile.cpp
+++ b/src/tests/HDFFile.cpp
@@ -24,7 +24,11 @@ using std::chrono::milliseconds;
 using std::chrono::steady_clock;
 using nlohmann::json;
 
-MainOpt getTestOptions() { return MainOpt(); }
+MainOpt getTestOptions() {
+  MainOpt mo;
+  mo.init();
+  return mo;
+}
 
 void merge_config_into_main_opt(MainOpt &main_opt, string JSONString) {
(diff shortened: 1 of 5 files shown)


src/tests/HDFFile.cpp
```diff
@@ -24,7 +24,11 @@ using std::chrono::milliseconds;
 using std::chrono::steady_clock;
 using nlohmann::json;
 
-MainOpt getTestOptions() { return MainOpt(); }
+MainOpt getTestOptions() {
+  MainOpt mo;
```
matthew-d-jones: Please use LLVM naming: upper CamelCase for variables, perhaps `TestOptions` or something like that.

Issue #217, opened 2018-07-02
SWMR cannot be disabled


matthew-d-jones:
The default value of the "use SWMR" flag is what is used, even if `"use_hdf_swmr": false` is used in the "new" command.

The default value is set here:
https://github.com/ess-dmsc/kafka-to-nexus/blob/37a879f5d67eebbb1fe3a718db64fe67e8294ad5/src/FileWriterTask.h#L44

The filewriter does hit this line:
https://github.com/ess-dmsc/kafka-to-nexus/blob/37a879f5d67eebbb1fe3a718db64fe67e8294ad5/src/CommandHandler.cpp#L225

So I would guess that reopening the HDF file (a couple of lines further down) is not resetting SWMR file writing.
Note, my test of whether SWMR was used or not is to try to open the file with HDF5 v1.8. Therefore another possibility is that reopening the file does not reset the block indexing method, so maybe SWMR does get disabled but the file still cannot be read by v1.8.

Mantid and also h5py (when installed via pip) are still on v1.8.

matthew-d-jones:
@dominikwerder @michele-brambilla 

For now I'll make a PR to temporarily set the default to disable SWMR so that I'm not blocked on a high priority ticket (involving testing the file writer with full ESS-style NeXus files).

dominikwerder:
Fixed in #227

PR #218, opened 2018-07-02
Change default UseHDFSWMR value to false

matthew-d-jones:
### Description of work

This sets the default back to not using SWMR, as there is a problem with the current implementation which is blocking other important tickets, see #217 .

No change to the documentation was necessary, it already implies that this is the default.

---

---


dominikwerder:
Fixed in #227

PR #238, opened 2018-08-01
Changed to dealing with fmt as a library and fixed bugs.

### LINKED ISSUE #157: The FMT library should be treated as a library
Currently, the FMT formatting library source is being compiled as a part of the "kafka_to_nexus__objects" library. Seeing as the conan package for FMT actually provides a binary for FMT, that should probably be used instead.

SkyToGround:
### Description of work

As the title says. I had to make some minor changes to the cmake-code and I found two build issues. The bonus is that we are now using a newer version of the fmt library. 

### Issue

Closes #157 

### Acceptance Criteria

Review the changes to see if they look reasonable.

### Unit Tests

N/A

### Other

We are now using the bincrafters conan package instead as it has the latest version of fmt and it provides its own _FindFmt.cmake_-file.

---

---


SkyToGround:
Looks like we have some build issues that I have to sort out before we can merge.

matthew-d-jones:
@SkyToGround I think you need to bump `cmake_minimum_required` up to 3.1 to use that syntax.

SkyToGround:
This was harder than expected. I will postpone this fix until @amues is back.

matthew-d-jones:
@SkyToGround I don't think we do that anywhere, the problem is probably just that you need to `cmake3` on CentOS. For the other platforms it looks like you are trying to use an option that the recipe doesn't have?

SkyToGround:
@matthew-d-jones It is now building properly. I have added cmake 3.10 as one of the required conan packages. What I could not remember yesterday was that the proper environment variables has to be set for the conan provided version of cmake to be invoked. This is done by running _activate_run.sh_ before running _cmake_. Because of old versions of conan on three of the docker images used by this project, I could not use the latest version of fmt and instead went for fmt 4.1.0.

mattclarke:
:+1: Looks good to me

PR #244, opened 2018-08-10
221 reject write message if job id in use

### LINKED ISSUE #221: Incorrect behaviour if a busy file writer is asked to create another file
If a file writer is already writing a file and it receives a message to create a new file with a different file name then it appears to give up on writing the first file and creates a new file. We need to decide what the correct behaviour is in this circumstance.

The behaviour should also be checked in the case that a different job id is used.

michele-brambilla:
### Description of work

Each writing task must be tagged with a unique ID. So far a bug allowed two different task to have the same ID, causing different misbehaviours. In this PR we introduce a fix that reject the "write new" command if the jobID is already in use.

### Issue

Closes #221 

### Acceptance Criteria

- the file writer should accept "write new" commands with different jobID
- if a "write new" command is received but the jobID is already in use should drop the command

---

---


src/CommandHandler.cpp
```diff
@@ -191,6 +191,12 @@ void CommandHandler::handleNew(std::string const &Command) {
       logMissingKey("job_id", Doc.dump());
       return;
     }
+    if (MasterPtr) { // workaround to prevent seg fault in tests
```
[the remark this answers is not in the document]
michele-brambilla: The Master interface sounds attractive, though ``getStreamMasterForJobID`` returns the reference to a ``StreamMaster`` pointer. In order this not tu be ``null`` we need to properly construct the StreamMaster, which in turn requires a number of things to be created. Could we then consider it a "unit test"? Anyway, I'll give it a try.

src/CommandHandler.cpp
```diff
@@ -414,7 +421,15 @@ void CommandHandler::handle(std::string const &Command) {
   if (auto x = find<std::string>("cmd", Doc)) {
     std::string CommandMain = x.inner();
     if (CommandMain == "FileWriter_new") {
-      handleNew(Command);
+      try {
+        handleNew(Command);
+      } catch (JobIdException &e) {
+        LOG(Sev::Critical, "{}", e.what());
```
dominikwerder: Should it be Critical?

src/CommandHandler.cpp
```diff
@@ -414,7 +421,15 @@ void CommandHandler::handle(std::string const &Command) {
...
+        throw JobIdException(e.what());
+      } catch (std::runtime_error &e) {
+        throw std::runtime_error(e.what());
```
dominikwerder: What's the purpose here? Could instead simply not catch at all, or?

src/CommandHandler.cpp
```diff
@@ -414,7 +421,15 @@ void CommandHandler::handle(std::string const &Command) {
   if (auto x = find<std::string>("cmd", Doc)) {
     std::string CommandMain = x.inner();
     if (CommandMain == "FileWriter_new") {
-      handleNew(Command);
+      try {
+        handleNew(Command);
+      } catch (JobIdException &e) {
+        LOG(Sev::Critical, "{}", e.what());
+        throw JobIdException(e.what());
```
dominikwerder: Why throw the same error message via a fresh constructed exception? Why not just `throw` if the same exception should bubble further up?

src/CommandHandler.h
```diff
@@ -7,6 +7,15 @@
 #include "Msg.h"
 #include <memory>
 
+// class JobIdException;
+class JobIdException : public std::runtime_error {
+public:
+  JobIdException(const std::string &What) : std::runtime_error(What) {}
+  JobIdException(const JobIdException &Other) = default;
```
dominikwerder: Do we need this constructor?

src/CommandHandler.h
```diff
@@ -7,6 +7,15 @@
...
+  JobIdException(JobIdException &&Other) = default;
```
dominikwerder: Also here, is this constructor used?

src/CommandHandler.h
```diff
@@ -7,6 +7,15 @@
...
+  ~JobIdException() = default;
```
dominikwerder: Not needed.

src/CommandHandler.cpp
michele-brambilla: As we discussed e can let  ``tryToHandle()`` (instead of ``handle(std::string const &)``) handle any exception that ``handleNew`` throws.

src/CommandHandler.h
```diff
@@ -7,6 +7,12 @@
 #include "Msg.h"
 #include <memory>
 
+// class JobIdException;
```
mattclarke: This comment just repeats the line below ;)
I think we can lose it

michele-brambilla: Right.. I wanted to add a couple of lines of documentation but I forgot. I'll an update soon.

PR #254, opened 2018-08-21
Catch exception from writer module init

### LINKED ISSUE #253: Catch exception during initialization of a writer module
Runtime errors during the initialization of a writer module should be catched.

dominikwerder:
### Description of work

- Catch exceptions from writer module initialization
- Add more details via `throw_with_nested` in some places
- Attempt HDF5 `flush` and `close` only given a valid file handle to avoid the corresponding exceptions in that case
- Factored body of loop into new function `extractStreamInformationFromJsonForSource`

### Issue

Closes #253 

### Acceptance Criteria

- Wrong settings especially in the `stream` children of `nexus_structure` should now be caught

### DIFF
--- a/src/HDFFile.cpp
+++ b/src/HDFFile.cpp
@@ -860,7 +860,7 @@ void HDFFile::close() {
       h5file.close();
       LOG(Sev::Debug, "closed");
     } else {
-      LOG(Sev::Debug, "File is not valid, skip close.");
+      LOG(Sev::Error, "File is not valid, skipping flush and close.");
     }
   } catch (std::exception &e) {
     LOG(Sev::Error, "ERROR could not close  file={}  trace:\n{}",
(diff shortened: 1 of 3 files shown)


src/HDFFile.cpp
```diff
@@ -853,11 +853,15 @@ void HDFFile::init(nlohmann::json const &nexus_structure,
...
-    h5file.close();
-    LOG(Sev::Debug, "closed");
+    if (h5file.is_valid()) {
+      LOG(Sev::Debug, "flushing");
+      flush();
+      LOG(Sev::Debug, "closing");
+      h5file.close();
+      LOG(Sev::Debug, "closed");
+    } else {
+      LOG(Sev::Debug, "File is not valid, skip close.");
```
matthew-d-jones: Sounds worthy of `Sev::Error` to me.

PR #262, opened 2018-08-27
Improved CMake code (JIRA issue DM-1019)

SkyToGround:
### Description of work

Minor changes to the CMake-code in order to modularise it and make future file-writing-module development slightly easier. 

### Issue

See JIRA issue DM-1019. This pull request does not close the JIRA issue but it does update it.

### Acceptance Criteria

Review the CMake code in order to determine if the changes are useful/reasonable.

### Unit Tests

N/A

### Other

N/A

---

## Code Review (To be filled in by the reviewer only)

- [ ] Is the code of an acceptable quality?
- [ ] Do the changes function as described and is it robust?

---

## Nominate for Group Code Review (Anyone can nominate it)
Indicate if you think the code should be reviewed in a Thursday code review session.

- [ ] Recommend for group code review

Also, nominate it on the code_review Slack channel (does someone want to automate this?).

PR #267, opened 2018-09-04
Register modules not globally

### LINKED ISSUE #265: PR 262 breaks writer module registration
PR #262 breaks the registration of file writer modules.

dominikwerder:
### Description of work

- Register writer modules in `MainOpt` instead of a global and pass map of writer modules as function parameters.
- Fixes the issues introduced by PR #262 
- Needs some more polish after this initial fix (#268)

### Issue

Closes #265 

### Acceptance Criteria

Filewriter should find its writer modules again and write files.

### Unit Tests

Existing integration tests.


michele-brambilla:
Stupid question: are this PR and #269 working together or in conflict? They both close the same issue.

dominikwerder:
Both fix #265 among other things, and they are not really in a logical conflict.

This PR (#267) not only closes #265 but is also a step towards a cleaner design where writer modules are no longer registered in a static global variable.

PR (#269) makes changes to the build process and moves again away from the intermediate cmake `static` library back to the original cmake `object` type library. This also fixes #265, but may fix as well some issues with the code coverage, so @SkyToGround may wants that anyway?

SkyToGround:
I think the use of globally registered modules is problematic, especially when writing unit tests. I also think that the ability to register a module using only two lines in the actual module is very nice. Thus I suggest that we use both solutions.

1. Keep the registration system as is.
2. Keep the newly developed system for passing around module information.
3. Instead of manually adding modules as you have done in _MaintOpt.cpp_, populate the `std::map` instances defined in `MainOpt.h` using the global registries.

During todays meeting at the DMSC, there were also suggestions for how to make registration easier safer and more reliable by using macros.

mattclarke:
I'm closing this as the initial issue was fixed in another PR. But the general concept is good, so I have created a ticket for that https://jira.esss.lu.se/browse/DM-1257

PR #277, opened 2018-09-18
Event logger

michele-brambilla:
### Description of work

Notify relevant events in the status topic that NICOS can interpret and report to the user. The log codes **must** match what NICOS expect.

### Acceptance Criteria

The log events must be when they are expected correctly and with the correct code. The `logEvent` function must be accessible from any component of the file writer.

### Unit Tests

Add 'EventLogger' test to check the fields in the log message and the log codes.

---

---

### DIFF
--- a/src/EventLogger.h
+++ b/src/EventLogger.h
@@ -7,23 +7,27 @@ class ProducerTopic;
 }
 
 namespace FileWriter {
+enum class StatusCode : int {
+  Start = 0,
+  Close = 1,
+  Error = 2,
+  Fail = 3,
+};
+std::string convertStatusCodeToString(StatusCode Code);
+
+void logEvent(std::shared_ptr<KafkaW::ProducerTopic> p, StatusCode Code,
+              const std::string &ServiceId, const std::string &JobId,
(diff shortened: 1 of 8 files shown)


src/EventLogger.h
```diff
@@ -0,0 +1,99 @@
...
+/// pair of service and job id.
+///
+class EventLogger {
+public:
+  template <class ProducerType>
+  void create(std::shared_ptr<ProducerType> Producer,
+              const std::string &ServiceId, const std::string &JobId) {
+    using namespace std::placeholders;
+    doLogEvent =
+        std::bind(logEvent<ProducerType>, Producer, _1, ServiceId, JobId, _2);
```
dominikwerder: Can be written in a more idiomatic way by using some `EventLoggerI` and imlementing a concrete EventLogger for that `ProducerType` and keeping the `Producer` itself as member.

michele-brambilla: As we discuss the class is not necessary. The `Steramer`, which is the one that requires it, return any error to the `StreamMaster`, which has all the knowledge required to report the error.

PR #279, opened 2018-09-20
Messages for ECP

dominikwerder:
### Description of work

Inform ECP of important events:

- Emit "Start" status message on file open
- Emit "Close" status message on successful file close
- Emit "Fail" if file can not be created (wrong json, I/O)

### DIFF
--- a/src/CommandHandler.cpp
+++ b/src/CommandHandler.cpp
@@ -17,7 +17,7 @@ namespace FileWriter {
 
 using nlohmann::json;
 
-static json parseOrThrow(std::string const &Command) {
+json parseOrThrow(std::string const &Command) {
   try {
     return json::parse(Command);
   } catch (json::parse_error const &E) {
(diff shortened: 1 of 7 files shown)


src/CommandHandler.cpp
```diff
@@ -453,39 +454,49 @@ void CommandHandler::handle(std::string const &Command) {
...
+  }
+  return StrS.str();
+}
+
+std::string format_nested_exception(std::exception const &E) {
+  std::stringstream StrS;
+  return format_nested_exception(E, StrS, 0);
+}
+
+static void dummy_emit_msg(std::string JobID, std::string Code,
```
michele-brambilla: Is this useful?

dominikwerder: Are you sure that you're looking at the correct commits?

6d4f4a1 does not contain that function anymore.

dominikwerder:
@michele-brambilla If the `StreamMaster` needs an improved interface for getting passed the `ServiceID` then I think we should get this merged, and address the passing of `ServiceID` to `StreamMaster` in a fresh PR, what do you think?

PR #294, opened 2018-09-27
Single option for kafka options 

### LINKED ISSUE #290: Change cli to use single option for kafka options
Means the user doens't have to worry about separating their kafka options

rerpha:
### Description of work

Replacing the two options for kafka options that are integers and strings to one single option. As CLI11 is now used no string formatting is needed to convert the integers to strings then sent off to kafka to set up brokers etc. 

### Issue

Closes #290 

### Acceptance Criteria

Tests pass 
I have tested this with an .ini file with a debugger and it parses correctly so I would assume it'd do the same for key value pairs in a command line call outside of an .ini file but haven't tested this yet. 

### Unit Tests

No tests modified. 

### Other

Updated readme to show example of new command line option. 

---

---


src/CLIOptions.cpp
```diff
@@ -146,13 +128,8 @@ void setCLIOptions(CLI::App &App, MainOpt &MainOptions) {
...
       "Stream-master option - topic write interval (milliseconds)");
-  addKafkaOption(
-      App, "-S,--kafka-config-strings",
-      MainOptions.StreamerConfiguration.Settings.ConfigurationStrings,
-      "LibRDKafka option (String value)");
-  addKafkaOption(
-      App, "-I,--kafka-config-ints",
-      MainOptions.StreamerConfiguration.Settings.ConfigurationIntegers,
-      "LibRDKafka option (Integer value)");
+  addKafkaOption(App, "-S,--kafka-config-options",
```
matthew-d-jones: Let's drop the `-options` I think `--kafka-config` is sufficient.

src/KafkaW/BrokerSettings.cpp
```diff
@@ -5,36 +5,9 @@
...
-    auto s1 = fmt::format("{:d}", c.second);
-    LOG(Sev::Debug, "set config: {} = {}", c.first, s1);
-    if (RD_KAFKA_CONF_OK !=
-        rd_kafka_conf_set(RdKafkaConfiguration, c.first.c_str(), s1.c_str(),
-                          ErrorString.data(), ErrorString.size())) {
-      LOG(Sev::Warning, "error setting config: {} = {}", c.first, s1);
-    }
-  }
-  for (auto &c : ConfigurationStrings) {
+  for (auto &c : KafkaConfiguration) {
```
matthew-d-jones: Can also be `const`

PR #300, opened 2018-10-10
Disable system tests because of build server issues

dominikwerder:
### Description of work

Disable some tests because there is some error in the build server setup.


mattclarke:
I am not sure disabling tests because there is a "infrastructure" issue is sensible.
It is still possible to merge even if the build fails.

mattclarke:
They still work locally, but do what you think is best

PR #303, opened 2018-10-10
Update jenkinsfile

amues:
### Description of work

Update Jenkinsfile to parallelise build and adopt the ECDC Jenkins shared library.

### Issue

N/A

### Acceptance Criteria

Functionality should not change.

### Unit Tests

N/A

### Other

N/A

---

---

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,10 @@ container_build_nodes = [
   'ubuntu1804': new ContainerBuildNode('essdmscdm/ubuntu18.04-build-node:1.2.0', 'bash -e')
 ]
 
+// Define number of old builds to keep. These numbers are somewhat arbitrary,
+// but based on the fact that for the master branch we want to have a certain
+// number of old builds available, while for the other branches we want to be
+// able to deploy easily without using too much disk space.
 def num_artifacts_to_keep
 if (env.BRANCH_NAME == 'master') {
   num_artifacts_to_keep = '5'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,394 +1,303 @@
...
+          }  // stage
+          stage("System tests: Run") {
+            sh """cd system-tests/
+            scl enable rh-python35 -- python -m pytest -s --junitxml=./SystemTestsOutput.xml .
+            """
+            junit "system-tests/SystemTestsOutput.xml"
+          }  // stage
+        } finally {
+          stage ("System tests: Clean Up") {
+            sh """rm -rf system-tests/output-files/* || true
```
rerpha: This my fault for not doing it, but whilst you're here would you mind sticking a comment in here that says something like `the statements below return true because the build should pass even if there are no docker containers or output files to be removed` as it's probably not clear why there is a `|| true` on the end of the statements

amues: Sure, I'll add it.

Jenkinsfile
```diff
@@ -1,394 +1,303 @@
...
 
-base_container_name = "${project}-${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
-
-def Object container_name(image_key) {
-    return "${base_container_name}-${image_key}"
+def num_artifacts_to_keep
+if (env.BRANCH_NAME == 'master') {
+  num_artifacts_to_keep = '5'
+} else {
+  num_artifacts_to_keep = '1'
```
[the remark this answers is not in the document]
amues: They are quite arbitrary, I'll add a comment.

dominikwerder:
Not sure if this is related to the changes in this PR, but I see that the build server now builds `boost` from source, which I think was not the case before, and takes quite some time. Or is this only because the pre-built `boost` conan package is currently not available for some reason? (ping also @matthew-d-jones )

amues:
I saw this happening for Ubuntu — that should not be happening, I'll look into this.

PR #306, opened 2018-10-11
Adding custom nexus file for system tests

### LINKED ISSUE #283: Create custom NeXus file for system tests
The default one the nexus streamer is using is huge and contains a lot of data which won't be used for testing. We should cut this down so everything gets streamed and the tests can run more quickly.

### LINKED ISSUE #301: Fix and re-introduce system tests
System tests are disabled in https://github.com/ess-dmsc/kafka-to-nexus/pull/300 because they fail 100% because of build server issues. Build servers need to get fixed, then these tests have to get re-enabled.

rerpha:
### Description of work

Ignore the lines added, it's just the json command message. It is needed for nexus geometry. 

Adding a custom-made nexus file based on the SANS2D_test.nxs file previously used. This has been reduced from ~100Mb to <2Mb so streamed data can be streamed fully into kafka before the tests finish. This means the tests will be more robust and will take less time to complete. 

### Issue

Closes #283 
Closes #301 

### Acceptance Criteria

System tests on jenkins should pass. if not the `sleep()` calls are probably not allowing enough time for data to be streamed/configuration to be received by the file-writer. 

---

---


rerpha:
please don't merge until after the maintenance window, the system tests are not currently running on jenkins. I will edit the jenkinsfile to run them next week when the maintenance is done if that's OK with you @amues

PR #321, opened 2018-10-30
Log filename and minor fixes

mattclarke:
### Description of work
Some minor fixes such as doxygen formatting, make_unique and explicit.

Also, it logs the name of the file being written.

### Issue
N/A

### Acceptance Criteria
Builds correctly and the changes are sensible.

### Unit Tests

N/A

### Other

N/A

---

---

### DIFF
--- a/src/CommandHandler.cpp
+++ b/src/CommandHandler.cpp
@@ -202,8 +202,7 @@ void CommandHandler::handleNew(std::string const &Command) {
   if (MasterPtr) {
     StatusProducer = MasterPtr->getStatusProducer();
   }
-  auto Task = std::unique_ptr<FileWriterTask>(
-      new FileWriterTask(Config.service_id, StatusProducer));
+  auto Task = std::make_unique<FileWriterTask>(Config.service_id, StatusProducer);
   if (auto x = find<std::string>("job_id", Doc)) {
     std::string JobID = x.inner();
     if (JobID.empty()) {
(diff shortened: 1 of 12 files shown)


src/kafka-to-nexus.cpp
```diff
@@ -24,12 +24,12 @@ void signal_handler(int Signal) {
 int main(int argc, char **argv) {
 
   fmt::print("kafka-to-nexus {:.7} (ESS, BrightnESS)\n"
```
rerpha: Not entirely related to this PR but since there are changes here, this `fmt::print` should be put in the `CLI::App` constructor, no point in having a separate print statement when the constructor takes a string and prints it. 

you can use `fmt::format()` to do this

src/FileWriterTask.cpp
```diff
@@ -98,6 +98,7 @@ void FileWriterTask::hdf_init(std::string const &NexusStructure,
   auto ConfigFileJson = hdf_parse(ConfigFile);
 
   try {
+    LOG(Sev::Info, "Creating HDF file {}", filename_full);
```
rerpha: would be nice to have `filename_full` LLVM'ed

src/KafkaW/Consumer.h
```diff
@@ -23,7 +23,7 @@ class ConsumerInterface {
 
 class Consumer : public ConsumerInterface {
 public:
-  Consumer(BrokerSettings opt);
+  explicit Consumer(BrokerSettings opt);
```
rerpha: can we rename opt to something a bit more descriptive?

src/CommandHandler.cpp
```diff
@@ -40,28 +39,9 @@ static void throwMissingKey(std::string const &Key,
 // In the future, want to handle many, but not right now.
 static int g_N_HANDLED = 0;
```
rerpha: let's just remove this as it's not being used anywhere outside of this class so really has no benefit of being there

src/CommandHandler.cpp
```diff
@@ -202,8 +182,8 @@ void CommandHandler::handleNew(std::string const &Command) {
   if (MasterPtr) {
```
SkyToGround: We have discussed this before but I want to re-iterate that I very much dislike implicit conversion of pointers to bool. I very much prefer to do `MasterPtr != nullptr`.

src/CommandHandler.cpp
```diff
@@ -288,9 +268,9 @@ void CommandHandler::handleNew(std::string const &Command) {
...
```
SkyToGround: Ditto.

rerpha:
I will make tickets for these comments.

PR #326, opened 2018-10-31
Simplify data gather

dominikwerder:
### Description of work

Reduce duplication of code: To prepare writing to HDF, we often have to gather n-dimensional json arrays into a buffer suitable for the call to h5cpp. Previously, we had some code duplication for the cases of numerical and string data. This is now unified.

### Acceptance Criteria

Tests should pass as before.

### DIFF
--- a/src/HDFFile.cpp
+++ b/src/HDFFile.cpp
@@ -66,6 +66,8 @@ private:
   size_t Size = 0;
 };
 
+static size_t const MAX_DIMENSIONS_OF_ARRAY = 10;
+
 template <typename DT>
 static std::vector<DT> populateBlob(nlohmann::json const &Value,
                                     size_t GoalSize) {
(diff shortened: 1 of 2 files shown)


src/HDFFile.cpp
```diff
@@ -39,63 +39,104 @@ static void writeAttribute(hdf5::node::Node &Node, const std::string &Name,
...
+    Buffer.push_back(Value);
+  }
+};
+
+class FixedStringItemHandler {
+public:
+  using DataType = char;
+  static void append(std::vector<DataType> &Buffer, nlohmann::json const &Value,
+                     size_t const ItemLength = 0) {
+    if (ItemLength >= 1024 * 1024) {
```
michele-brambilla: Can you define this max length somewhere outside the function? Maybe put `MAX_DIMENSIONS_OF_ARRAY` and this together at the beginning of the file makes "less hidden" these magic numbers

PR #328, opened 2018-11-01
Improve logging

SkyToGround:
### Description of work

Seeing as the logging code for the file writer (and forwarder) is somewhat confusing and hard to modify, I decided to spend a few hours improving the situation. My goal was to keep and improve upon the features that the existing logging implementation already had without regressing in functionality. This is what I have done:

* Switched to the latest version of the graylog-logger library, made it required and updated the relevant CMake code.
* Removed existing console and file logging functionality and switched over to using the corresponding implementation in graylog-logger.
* Created a new kafka-to-graylog interface for the graylog-logger library to replace the existing kafka-to-graylog code.
* Removed the `enum class Sev` type and replaced it with `using Sev = Log::Severity`.
* Removed global variables defined in __logger.cpp__ and fixed the rest of the code base to work without those.
* Minor LLVM:ification of some of the options variable names.
* Added support for both textual and numeric selection of verbosity/logging level in the command line interface.

When looking at this pull request, there are several important things to consider:
* I found a serious issue in the existing code where log messages are self re-generating and thus are basically generated in an infinite loop. This should probably be fixed sooner rather than later.
* Although I am a big proponent of "profile first, optimise later", it might be that some of my (possibly) worst pieces of code (in terms of performance) should be optimised before a potential merge. See if you can find them.
* It might be appropriate to have some "trace" functionality for cases when you care about performance. I consider some of the `LOG()` calls wrapped in `if`-statements to be candidates for this.
* Is there some obvious and important functionality that was lost when I modified the code?
* I do not see it as a problem when making graylog-logger a required dependence for the file-writer (seeing as we are using Conan) but maybe there is a good argument against graylog-logger.

### Issue

n/a

### Unit Tests

n/a

---

---


mkoennecke:
It is not clear to me if the current functionality of logging to files and stdout is being implemented through the graylog library. This would be requirement for SINQ.   Does the graylog library allow for plugins to other logging systems? Though the future of logging for SINQ is not yet decided, it is unlikely that we decide for the graylog monster. Currently, we favor MongoDB. IMHO, logging should be implemented in such a way that backends are a plugin. Kafka-to-NeXus will be used not only at ESS but also at PSI and ISIS. Thus some flexibility is a requirement.

SkyToGround:
@mkoennecke The library that I put together (graylog-logger) does not by any means require that a Graylog server is available to be useful. I also implemented console and file logging into the library which is what is used here. Adding additional logging interfaces is relatively simple and this is what I did for the Kafka to Graylog functionality that already existed in the `logger.h`/`logger.cpp` files.

I fully admit that the logging library that I wrote has a lot of shortcomings and might not be something that we should use going forward but I consider the implementation in this pull request to be superior in lines of code, functionality and simplicity (discoverability) to the old one.

src/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ find_package(HDF5_SIMPLE REQUIRED)  # the built in FindHDF5 does not handle 1.10
 find_package(FlatBuffers REQUIRED)
 find_package(h5cpp REQUIRED)
 find_package(StreamingDataTypes COMPONENTS 2246f84b16423b7ac18e901bf4c43a006fe635dc)
-find_package(GraylogLogger)
+find_package(GraylogLogger REQUIRED)
```
dominikwerder: This is a regression, among others because it introduces way tighter **coupling** than necessary.
It would be a better design if the different logging endpoints (like graylog, kafka-gelf, …) could be chosen optionally, while the filewriter only provides the common abstractions which are used for interfacing with the logging endpoints.
`GraylogLogger` also depends on `asio` which is another dependency that could be avoided by a more **modular** approach.

src/KafkaGraylogInterface.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  auto Producer = std::make_shared<KafkaW::Producer>(BrokerSettings);
+  auto Topic = std::make_shared<KafkaW::Producer::Topic>(Producer, TopicName);
+  Topic->enableCopy();
+
+  int MessageTimeoutMS = 50;
+  while (true) {
+    bool GotMessage = MessageQueue.time_out_pop(TmpMsg, MessageTimeoutMS);
+    if (GotMessage) {
+      if (std::string("exit") == TmpMsg.MessageString and
```
dominikwerder: Magic value `"exit"` as a token for thread exit is bad design because it is brittle. Use a separate type.

src/KafkaW/Producer.cpp
```diff
@@ -170,9 +171,16 @@ void Producer::poll() {
...
-  }
+
+  //  char *bp;
+  //  size_t size;
+  //  FILE *stream;
+  //  stream = open_memstream(&bp, &size);
+  //  rd_kafka_dump(stream, RdKafkaPtr);
+  //  LOG(Sev::Debug, "{}", bp);
+  //  fclose(stream);
+  //  free(bp);
```
dominikwerder: Commented code.

zjttoefs:
Just to understand: Are you not using Graylog? 
How much is running without a logger implementation an actual use case?

SkyToGround:
> It was/is clear that a common interface for log endpoints is required to support different logging systems in a decoupled and modular way.

> Unfortunately, this PR falls short of that goal, because it uses as a common interface a class from the graylog-logger library which in turn also provides concrete implementations e.g. to Graylog and unconditionally pulls in networking code and a dependency on asio which is an unnecessary liability if one does not use Graylog.

The logging can in principle be decoupled further than what was done in this pull-request. With that said, the point of contact (in the C++ code) is currently the `setupLogging()` function and the `#define LOG` macro. In comparison, the system that currently exists in the master branch of the file-writer has the following points of contact:

* The `#define LOG` macro.
* A set-up function (`setupLoggerFromOptions()`) that is defined outside the logging interface and that uses (and can thus modify) a `struct` that is used across the code-base.
* The existance and use of globals (`int log_level` and `std::string g_ServiceID`).
* Use of non-standard log-levels.

Thus the logging implementation in this pull request **has less coupling** with the rest of the code base than the old version.

Though I generally agree that we should try to (at least some extent) limit the number of dependencies we have, the notion that the _asio_ library is a liability is wrong when comparing it to other dependencies of the file writer.

_asio_ has become the de-facto C++ way of doing network (socket) programming to the extent that it appears very likely that [it will be included in C++20](https://isocpp.org/std/status).
In comparison, we have several dependencies in the file-writer that could prove to be much more of a liability, for example: _CLI11_, _JSON for modern C++_ and _Date_. The first two have both been created by single developers and replacing them will require a not insignificant development effort whereas the last one is a library which has been custom modified for use with the file-writer.

> A more modular design would be to let the filewriter use a common abstraction for the logging handlers. That abstraction should not automatically pull in dependencies or concrete implementations. On top of that implement the concrete logging handlers which can then optionally be used by the filewriter.

As I have mentioned in my previous comments and the original pull-request description, this pull-request is NOT the last word in logging. But it is _definitively an improvment_ over the current implementation and until a better one comes along, I suggest that we use this one.

SkyToGround:
@dominikwerder By the way, I agree with your comments on the code but I want someone else to take a look at pull request before I start modifying it.

dominikwerder:
@zjttoefs Yes, we use mostly the console log output. There's also already the option to write log messages as GELF to a Kafka topic (since we use `librdkafka` anyway). Graylog in turn can read GELF from Kafka out of the box.

dominikwerder:
> Thus the logging implementation in this pull request **has less coupling** with the rest of the code base than the old version.

I do not even attempt to measure an overall "more" or "less". The PR decreases coupling in certain areas :+1: but also introduces coupling in areas where it could be avoided, so if we're already touching that part of code, I feel that we could use the opportunity to factor it a bit further.

> Though I generally agree that we should try to (at least some extent) limit the number of dependencies we have, the notion that the _asio_ library is a liability is wrong when comparing it to other dependencies of the file writer.

I agree that `asio` is high quality. Difference is that we would only call into it when using *that* specific logging endpoint.

matthew-d-jones:
@dominikwerder So you are concerned about the 500kb that asio adds?

dominikwerder:
@matthew-d-jones Any dependency can break or fail to build in the future on some new system/version that one wants to support. I hope we agree that minimizing the number of **unused** dependencies by making them optional is in general not a bad thing?

dominikwerder:
Splitting this up into interface plus log endpoints is just a suggestion from my side. Maybe there a reasons why splitting up into interface plus optional log endpoint implementations is too difficult, or not worth it?

matthew-d-jones:
@dominikwerder Sure, I don't disagree with that. It is just a case of weighing the risk of asio causing a problem with the convenience of graylog-logger providing the interface for the all the log methods we need for several codebases.

matthew-d-jones:
@SkyToGround Would it be straightforward to make graylog-logger buildable without asio?
Adding an option to the conan package, which in turn sets a cmake variable when graylog-logger is built is straightforward.

SkyToGround:
@matthew-d-jones Yes, it would be relatively easy.

matthew-d-jones:
hm, I'm inclined to agree with Mark and Dominik on this one, and am tempted to resurrect this idea: https://github.com/ess-dmsc/graylog-logger/issues/11 (to implement a [spdlog](https://github.com/gabime/spdlog) plugin using graylog-logger). spdlog is a very popular and performance-proven logging library with "plugin" capability and uses `fmt` library underneath, so is very compatible with the current graylog-logger interface.
I believe this would solve the concerns on this PR, would be easy to achieve and could volunteer Michal's time to implement it. Let me know what you think.

mattclarke:
Won't be merged but will be used as the basis for https://jira.esss.lu.se/browse/DM-1168

PR #342, opened 2018-11-14
Removing Pollstatus

### LINKED ISSUE #339: Remove PollStatus and replace with an enum in Msg.h
Pollstatus is a useless wrapper for an Enum that it wraps. 

Instead of using Pollstatus, add an enum as a public member variable in Msg.h (or at least a getter/setter function set) 

Any places where Pollstatus is used should just instead use a unique pointer to a `KafkaW::Msg`

rerpha:
### Description of work

Removing Pollstatus from KafkaW as it is a useless wrapper of an `enum` 

This will simplify the task of switching from the C to the C++ API for rdkafka.

### Issue

Closes #339 

### Acceptance Criteria

Msg.h now contains a pollstatus enum and poll() methods return a unique pointer to the `msg` class that wraps the data from kafka. 

### Unit Tests

No tests modified. 
---

---


rerpha:
Still need to do modify tests.

src/KafkaW/Consumer.cpp
```diff
@@ -235,30 +235,22 @@ void Consumer::dumpCurrentSubscription() {
...
   }
 
   static_assert(sizeof(char) == 1, "Failed: sizeof(char) == 1");
-  std::unique_ptr<Msg> m2 = std::make_unique<Msg>(
-      (std::uint8_t *)msg->payload, msg->len,
-      [msg]() { rd_kafka_message_destroy(msg); }, msg->offset);
   if (msg->err == RD_KAFKA_RESP_ERR_NO_ERROR) {
-    return PollStatus::newWithMsg(std::move(m2));
+    return std::make_unique<Msg>((std::uint8_t *)msg->payload, msg->len,
+                                 [msg]() { rd_kafka_message_destroy(msg); },
```
SkyToGround: 👍

src/Master.cpp
```diff
@@ -91,9 +91,9 @@ void Master::run() {
   while (do_run) {
     LOG(Sev::Debug, "Master poll");
     auto p = command_listener.poll();
-    if (auto msg = p.isMsg()) {
+    if (p->getStatus() == KafkaW::PollStatus::Msg) {
```
SkyToGround: 👍

src/Master.cpp
```diff
@@ -91,9 +91,9 @@ void Master::run() {
   while (do_run) {
     LOG(Sev::Debug, "Master poll");
     auto p = command_listener.poll();
```
SkyToGround: I dislike the use of single character variable names like this one. Maybe do something like this instead:
```suggestion
    auto PollResult = command_listener.poll();
```

src/Master.cpp
SkyToGround:
```suggestion
    if (PollResult->getStatus() == KafkaW::PollStatus::Msg) {
```

src/Master.cpp
```diff
@@ -91,9 +91,9 @@ void Master::run() {
...
-    if (auto msg = p.isMsg()) {
+    if (p->getStatus() == KafkaW::PollStatus::Msg) {
       LOG(Sev::Debug, "Handle a command");
-      this->handle_command_message(std::move(msg));
+      this->handle_command_message(std::move(p));
```
SkyToGround:
```suggestion
      this->handle_command_message(std::move(PollResult));
```

PR #350, opened 2018-11-20
Fix bug in queryNumberOfPartitions

### LINKED ISSUE #349: Wrong return value in queryNumberOfPartitions
`queryNumberOfPartitions` returns 1 even in the case of failure.

dominikwerder:
### Description of work

Closes #349


rerpha:
This throws an error in the system tests when trying to find the topic `TEST_writerCommand` 

I have looked on the kafka instance that the system tests run and the topic is there, so it shouldn't be throwing an exception

dominikwerder:
@rerpha I can currently not build `master` on my machine because of some conan `date` lib build fail, hope I can reproduce this tomorrow.

PR #353, opened 2018-11-20
340 broker and topic settings as structs

### LINKED ISSUE #340: Convert BrokerSettings + TopicSettings to structs
everything is already public so we can just change the declaration. 

The configuration should also only be strings so only one Map is needed for this as well in TopicSettings 

Also, in the C++ API, the configuration/settings for topics may well have a parent class, and if this is the case these two structs/classes could inherit from a common interface.

mchorazak:
### Description of work

- Both broker and topic settings are now structs.
- Because C++ api indeed has a common interface for configuration/settings I created a
SettingsInterface.h to reflect this relation. 
- Currently TopicSettings are not explicitly set anywhere but `apply(rd_kafka_topic_conf_t *conf)` method is called and removing entire struct would require changes beyond kafkaW. If we want to consider it I'd suggest discussing it and creating a separate ticket.

### Issue

Closes #340 

### Acceptance Criteria
- BrokerSettings.h/cpp
- TopicSettings.h/cpp
- SettingsInterface.h

Functionality itself does not change
### Unit Tests

n/a

### Other

n/a

---

---

### DIFF
--- a/src/KafkaW/BrokerSettings.h
+++ b/src/KafkaW/BrokerSettings.h
@@ -1,7 +1,6 @@
 #pragma once
 
-#include <map>
-#include <string>
+#include "SettingsInterface.h"
 
 struct rd_kafka_conf_s;
 typedef struct rd_kafka_conf_s rd_kafka_conf_t;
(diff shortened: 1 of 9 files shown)


src/KafkaW/TopicSettings.h
```diff
@@ -1,17 +1,15 @@
...
-#include <string>
 
 namespace KafkaW {
 
-class TopicSettings {
-public:
+struct TopicSettings : public KafkaW::SettingsInterface {
   TopicSettings();
-  void applySettingsToRdKafkaConf(rd_kafka_topic_conf_t *conf);
+  virtual void apply() override{};
```
rerpha: This implementation doesn't do anything and isn't used. I suggest we get rid of `SettingsInterface` until we convert to the C++ API, as we can then use something like this for the interface: 

```
struct SettingsInterface {
  virtual void apply(RdKafka::Conf &KafkaConf) = 0;
  std::map<std::string, std::string> Configuration;
};

```

src/KafkaW/SettingsInterface.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <map>
+#include <string>
+namespace KafkaW {
+struct SettingsInterface {
```
rerpha: see above

src/KafkaW/BrokerSettings.h
```diff
@@ -1,21 +1,21 @@
...
 
 namespace KafkaW {
 
 /// Collect options used to connect to the broker.
-class BrokerSettings {
-public:
+struct BrokerSettings : public KafkaW::SettingsInterface {
   BrokerSettings() = default;
+
+  virtual void apply() override{};
```
rerpha: see above

PR #356, opened 2018-11-21
Fix Streamer/Kafka interaction

### LINKED ISSUE #349: Wrong return value in queryNumberOfPartitions

### LINKED ISSUE #355: Loop causes incorrect offset lookups
The loop [in Consumer::addTopic](https://github.com/ess-dmsc/kafka-to-nexus/commit/3aed9ccdbb27e424c52464c26d016c1334bb2e4e) causes `rd_kafka_offsets_for_times` to be executed for each partition, which in turn overwrites the results for the lookups of previous loop iterations with wrong values.

### LINKED ISSUE #358: UB in Streamer test
The test introduced [here](https://github.com/ess-dmsc/kafka-to-nexus/blame/12baeb406fe96d44ecb00967c3bc5d45dbb9223f/src/tests/StreamerTest.cpp#L52-L63) causes SEGV. 

It can obviously not work in the current form because it will lead to dereference of nullptr.

This bug only surfaces now, probably due to unrelated changes in other places as usual with undefined behavior.

I'm not very familiar with `Streamer` class, so I'm not sure what exactly we want to test here.

dominikwerder:
### Issue

Closes #349 
Closes #355 
Closes #358

### DIFF
--- a/src/KafkaW/Consumer.cpp
+++ b/src/KafkaW/Consumer.cpp
@@ -148,9 +148,19 @@ void Consumer::init() {
   PartitionList = rd_kafka_topic_partition_list_new(16);
 }
 
-void Consumer::addTopic(std::string Topic,
-                        const std::chrono::milliseconds &StartTime) {
+void Consumer::addTopic(std::string Topic) {
   LOG(Sev::Info, "Consumer::add_topic  {}", Topic);
+  int Error = rd_kafka_subscribe(RdKafka, PartitionList);
+  KERR(RdKafka, Error);
+  if (Error) {
+    throw std::runtime_error("can not subscribe");
+  }
+}
+
+void Consumer::addTopicAtTimestamp(std::string const Topic,
+                                   std::chrono::milliseconds const StartTime) {
+  LOG(Sev::Info, "Consumer::addTopicAtTimestamp  Topic: {}  StartTime: {}",
+      Topic, StartTime.count());
 
   auto numberOfPartitions = queryNumberOfPartitions(Topic);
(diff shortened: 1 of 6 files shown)


SkyToGround:
The removal of tests is not the solution here unless you can give a good reason. The file writer has too few tests as it is. Give me a while and I will take a look at it.

dominikwerder:
@SkyToGround I have removed the faulty test because I need to be able to run the test suite without SEGV, I hope you agree. I added a passing version of that test, but that commit is not yet pushed to this branch.

SkyToGround:
@dominikwerder  Well the best solution is to actually fix the test. The alternative is to disable the test temporarily by prepending `DISABLED_` to the name of the test.

I have to admit that the test was/is slightly fragile, which is one of the reasons that it crashed the executable. The other reason is because you increased the timeout from 1 second to 10 seconds.

I am going to see if there is some easy fix for this that does not also break the test.

dominikwerder:
@SkyToGround Like I said, I have already fixed it.

SkyToGround:
@dominikwerder Then I suggest that you push that fix to the branch?

dominikwerder:
@SkyToGround Yes, the dereference of `nullptr` caused SEGV.

The reassignment of `TestStreamer.ConsumerCreated` via `std::future::operator=()` that you mentioned is a small nuisance, but it passes.

dominikwerder:
Great that you see how to improve your tests, I saw them just after merge. I suggest you open a feature branch.
In this PR we just applied some fixes while keeping the changes as minimal as possible and making the changes to the tests in the same style as in your original.

PR #359, opened 2018-11-22
Reducing docker image size

### LINKED ISSUE #357: Reduce docker image
a lot of the RUN commands can be combined

rerpha:
### Description of work

I have reduced the docker image size for the filewriter by reducing the amount of layers in the image. 

I'm also looking at a really odd tool called docker-slim which seems to massively reduce the image size so might be worth having a look at this and testing with it.

### Issue

Closes #357 

### Acceptance Criteria

To test: run `docker build .` in the top directory and use `docker images` to compare size. 

I will push to docker hub if this gets merged. 

### Unit Tests

None

### Other

None

---

---


rerpha:
The build is failing due to no space left on the itest node. 

I think this is a bit trivial, and once compressed doesn't actually save much/any space at all. When building locally though it does save a bit of space so might still be useful.

PR #364, opened 2018-11-23
Move log and junit output archiving into finally step

### LINKED ISSUE #352: add finally() call in jenkinsfile to always archive sys test logs even if they fail
currently in the jenkinsfile the logs are being archived in a separate step of the system tests pipeline. Instead of this, wrap the pytest run shell script in a try/finally and archive the logs even if the tests fail.

rerpha:
### Description of work

This means logs and junit output should still be published when a system test fails.

### Issue

Closes #352 

### Acceptance Criteria

N/A

### Unit Tests

N/A

### Other

N/A

---

---

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -280,6 +280,7 @@ def get_system_tests_pipeline() {
             scl enable rh-python35 -- python -m pip install --user -r system-tests/requirements.txt
             """
           }  // stage
+          timeout(time: 30, activity: true)
           stage("System tests: Run") {
             // Stop and remove any containers that may have been from the job before,
             // i.e. if a Jenkins job has been aborted.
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -287,10 +287,12 @@ def get_system_tests_pipeline() {
             sh """cd system-tests/
             scl enable rh-python35 -- python -m pytest -s --junitxml=./SystemTestsOutput.xml .
             """
-            junit "system-tests/SystemTestsOutput.xml"
-            archiveArtifacts "system-tests/logs/*.log"
           }  // stage
         } finally {
+          stage("System tests: Archive") {
+            junit "system-tests/SystemTestsOutput.xml"
```
amues: As this is inside the `finally` block, it might be desirable to make this step fail-proof in the case test files are not created, to prevent the clean-up stage from being skipped. This can be done by changing this line to `junit allowEmptyResults: true, testResults: 'system-tests/SystemTestsOutput.xml'`. It has one disadvantage, though, as if the tests do not run because of a Jenkins job configuration issue (in which case files would not be generated), you might miss it. A simple workaround would be to move the clean-up stage to the start of the `finally` block.

Jenkinsfile
```diff
@@ -287,10 +287,12 @@ def get_system_tests_pipeline() {
...
+            archiveArtifacts "system-tests/logs/*.log"
```
amues: The same is valid here. To make this step ignore missing files, you can change the line to `archiveArtifacts allowEmptyArchive: true, artifacts: 'system-tests/logs/*.log'`. Feel free to choose the alternative you consider more adequate here.

rerpha:
@amues i have moved the cleanup stage before the archiving stage, I think we should disallow empty logs and empty test results as this indicates something went wrong in the tests, do you think that is reasonable?

PR #365, opened 2018-11-26
Streamer logs a warning if timestamp is missing

michele-brambilla:
### Description of work

With this PR an explicit check for a valid timestamp (_i.e. > 0_) in the flatbuffer message is checked. If the timestamp is not set the Streamer logs an error message and returns `ProcessMessageResult::ERR`.

### Issue

DM-1141

### Acceptance Criteria

*List the changes in functionality or code that the reviewer needs to review.*

### Unit Tests

* Add test for missing (_i.e. > 0_) timestamp field.
* Changed `StreamerTestDummyReader` class so that `timestamp(FlatbufferMessage const &Message)` returns a small, but valid value (it returns 1)

---

---

### DIFF
--- a/src/Streamer.cpp
+++ b/src/Streamer.cpp
@@ -151,7 +151,7 @@ FileWriter::Streamer::pollAndProcess(FileWriter::DemuxTopic &MessageProcessor) {
     return ProcessMessageResult::OK;
   }
 
-  if (static_cast<std::int64_t>(Message->getTimestamp()) == 0) {
+  if (Message->getTimestamp() == 0) {
     LOG(Sev::Error,
         "Message from topic \"{}\", source \"{}\" has no timestamp, ignoring",
         MessageProcessor.topic(), Message->getSourceName());
--- a/src/Streamer.cpp
+++ b/src/Streamer.cpp
@@ -151,6 +151,13 @@ FileWriter::Streamer::pollAndProcess(FileWriter::DemuxTopic &MessageProcessor) {
     return ProcessMessageResult::OK;
   }
 
+  if (static_cast<std::int64_t>(Message->getTimestamp()) == 0) {
+    LOG(Sev::Error,
+        "Message from topic \"{}\", source \"{}\" has no timestamp, ignoring",
+        MessageProcessor.topic(), Message->getSourceName());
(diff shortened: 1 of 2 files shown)


src/Streamer.cpp
```diff
@@ -151,6 +151,13 @@ FileWriter::Streamer::pollAndProcess(FileWriter::DemuxTopic &MessageProcessor) {
     return ProcessMessageResult::OK;
   }
 
+  if (static_cast<std::int64_t>(Message->getTimestamp()) == 0) {
```
dominikwerder: Cast is not needed here because compare with `0`.

src/tests/StreamerTest.cpp
```diff
@@ -228,6 +239,37 @@ class StreamerProcessTimingTest : public ::testing::Test {
   StreamerStandIn TestStreamer;
 };
 
+TEST_F(StreamerProcessTimingTest, MessageHasNoTimestamp) {
```
dominikwerder: Name of the test should also contain an indication of the expected behavior.

src/tests/StreamerTest.cpp
```diff
@@ -228,6 +239,37 @@ class StreamerProcessTimingTest : public ::testing::Test {
...
+  FlatbufferReaderRegistry::Registrar<StreamerNoTimestampTestDummyReader>
```
dominikwerder: Could that throw if schema id is already registered? Tests are run in undefined order in general.

src/tests/StreamerTest.cpp
mattclarke: :+1:

src/Streamer.cpp
michele-brambilla: :+1:

src/tests/StreamerTest.cpp
michele-brambilla: :+1:

src/tests/StreamerTest.cpp
michele-brambilla: Really don't think so (and I've never seen it throwing)
- the reader is registered only there
- other reader are registered multiple times in different units

mattclarke:
Looks okay to me

PR #369, opened 2018-11-28
Fix error in `commitOffsets`

michele-brambilla:
### Description of work

The `KafkaW::Consumer` throws an error if `rd_kafka_commit` returns something different from `RD_KAFKA_RESP_ERR_NO_ERROR`. This happens for example when there are no older messages (eventually within the specified time window) in the log. According to the discussion in this [ticket](https://github.com/confluentinc/confluent-kafka-python/issues/71) though

>   _NO_OFFSET is a "soft error", it means it didnt have to commit the offset because the current offset had already been committed. 

With this PR `RD_KAFKA_RESP_ERR__NO_OFFSET` is not considered an error, but a waning is emitted.

### Issue

[DM-1197](https://jira.esss.lu.se/browse/DM-1197)

### Other

Can someone please give it a try?

---

---


dominikwerder:
Is there a test case which demonstrates these possibilities, or how should one set up such a test?

PR #370, opened 2018-12-03
Refactoring URI class to comply with LLVM standards

rerpha:
### Description of work

Replace most of the functionality in URI with a regular expression which will find everything we need. 

Removing `scheme` as it isn't used anywhere. 

Will mirror this on the forwarder when it is ready 

### Issue

DM-1192

### Acceptance Criteria

Only change is removal of `is_alpha()` function and replacement with `std::string::find_first_of()` - same functionality just a single call 

### Unit Tests

Tests added for double colon in scheme and port 

Tests removed for scheme (scheme has now been removed) 

### Other

No documentation modified. 

---

---


src/URI.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+URI::URI(std::string URIString) { parse(URIString); }
+
+static vector<std::string> findProtocol(const std::string &URIString) {
+  auto Slashes = URIString.find("://");
+  if (Slashes == std::string::npos or Slashes == 0) {
+    return {std::string(), URIString};
+  }
+  auto Protocol = URIString.substr(0, Slashes);
+  if (Protocol.find_first_of("0123456789") != std::string::npos) {
```
mattclarke: This seems a bit brittle to me. What happens if someone accidentally does `URI.parse("http:://my.Host")`?
BTW: there should be a unit test for this.

I think using a simple regex would be a better option.

src/URI.cpp
```diff
@@ -25,100 +26,23 @@ static std::string findTopicFromPath(std::string URIString) {
...
-      } else {
-        return {URIString.substr(2, SlashPos - 2), std::string(),
-                URIString.substr(SlashPos)};
-      }
-    }
-  }
-}
+void URI::parse(std::string URIString) {
+  std::smatch Matches;
+  std::regex Regex(R"(\/\/([\w.]+):?(\d+)?([/\w-]+)?)");
```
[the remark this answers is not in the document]
rerpha: I had a look, but I can't figure out if we can use that to capture the port?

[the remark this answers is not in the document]
dominikwerder: I think that the regex from the RFC would yield host and port together in `$4` but that could probably be modified.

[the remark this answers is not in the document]
rerpha: Will have a play around with it.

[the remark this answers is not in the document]
rerpha: Have used a modified version that makes hosts mandatory and captures the port separately.

rerpha:
currently failing because topic is not being found

dominikwerder:
> currently failing because topic is not being found

(ping @michele-brambilla )

rerpha:
My own fault, sorry. the most recent commit should fix it

src/URI.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+void URI::UpdateHostPort() {
+  HostPort = Port != 0 ? fmt::format("{}:{}", Host, Port) : Host;
+}
+
+URI::URI(const std::string &URIString) { parse(URIString); }
+
+void URI::parse(const std::string &URIString) {
+  std::smatch Matches;
+  std::regex Regex(
```
mattclarke: Can we have a comment saying roughly what the regex is doing please?

src/URI.cpp
```diff
@@ -0,0 +1,30 @@
...
+      R"(\s*(([^:/?#]+):)?(//([^/?#:]+))(:(\d+))?/?([a-zA-Z0-9._-]+)?\s*)");
+  std::regex_match(URIString, Matches, Regex);
+  if (!Matches[4].matched) {
+    throw std::runtime_error("Host not found when trying to parse URI");
```
mattclarke: Can you check how this exception is handled please? Is it caught?

src/URI.h
```diff
@@ -6,53 +6,37 @@
...
   /// Parses the given `uri`
-  void parse(string uri);
-
-  /// Given a `http://www.example.com` scheme will contain `http`.
-  std::string scheme;
+  void parse(const std::string &URIString);
 
   /// Parsed hostname.
-  std::string host;
+  std::string Host;
```
mattclarke: Do we need both Host and HostPort?
Host seems to be used by the gelf stuff only. @dominikwerder could that take HostPort instead (assuming no port has been supplied, so it will still be the same value as Host)?

src/tests/URITests.cpp
```diff
@@ -0,0 +1,86 @@
...
+  ASSERT_EQ(TestURI.Port, (uint32_t)0);
+}
+
+TEST(URI, host_port) {
+  URI TestURI("//myhost:345");
+  ASSERT_EQ(TestURI.Host, "myhost");
+  ASSERT_EQ(TestURI.Port, (uint32_t)345);
+}
+
+TEST(URI, expectedURIStringWhenHostPortAndTopicAreSpecified) {
```
mattclarke: Change to snake case

src/tests/URITests.cpp
```diff
@@ -0,0 +1,86 @@
+#include "URI.h"
+#include <gtest/gtest.h>
+
+using namespace uri;
+
+TEST(URI, host) {
```
mattclarke: Better test names for some of the tests.
E.g. host_only_gives_host_plus_default_port

There are others below too

src/URI.h
dominikwerder: That should work.

PR #374, opened 2018-12-07
Update README on SWMR usage

dominikwerder:
### Description of work

Clarify that SWMR is used by default.

### Issue

Closes DM-1123

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,10 +288,21 @@ fails to initialise.  In order to abort writing of that file instead, set in the
 
 The filewriter can use HDF's Single Writer Multiple Reader feature (SWMR).
 
+SWMR-mode writing is enabled by default.
+
+To disable SWMR, add to the `FileWriter_new` command:
+
+```json
+{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -288,10 +288,21 @@ fails to initialise.  In order to abort writing of that file instead, set in the
...
+  "use_hdf_swmr": false
+}
+```
+
 To write, as well as to read HDF files which use the SWMR feature requires at
-least HDF version 1.10.  This means that also tools like HDFView are required
+least HDF5 version 1.10.  This means that also tools like HDFView are required
 to link against at least HDF 1.10 in order to open HDF files which were written
-in SWMR mode.
+in SWMR mode.  One can also use the HDF5 tool `h5repack` to convert the file
```
michele-brambilla: Does `h5repack` require some specific option? If so, can you add a simple example?

dominikwerder: The usage of `h5repack` is summarised in `h5repack --help`. Added comment to readme.

PR #386, opened 2019-01-11
DM-1174: fix streaming docs

mattclarke:
### Description of work
Some of the documentation was out of date, specifically relating to streams.
This PR fixes that and updates other parts of the commands.md file too.

Also, removed some legacy documentation.

### Issue

Closes DM-1174

### Acceptance Criteria

Make sure the commands.md file makes sense.

### Unit Tests

N/A

### Other

N/A

---

---


documentation/commands.md
```diff
@@ -1,233 +1,243 @@
...
-In the command, the `nexus_structure` defines the HDF hierarchy.
-The `nexus_structure` represents the HDF root object.  The following example
-shows how the HDF tree can be constructed using `children`.
-A child of type `stream` is a marker that a `HDFWriterModule` will insert
-the data into from a Kafka stream at that point in the hierarchy.  The options under
-the key `stream` specify the stream details containing, at least, the topic, the source name and
-the `HDFWriterModule` which should be used for writing.
-Depending on the `HDFWriterModule`, there will be more options specific to the
-`HDFWriterModule`.
+Commands are generally send through Kafka via the broker and topic specified by the
```
SkyToGround:
```suggestion
Commands are generally sent through Kafka via the broker and topic specified by the
```

documentation/commands.md
```diff
@@ -1,233 +1,243 @@
...
 
-- [Groups](docs/groups.md)
-- [~~Datasets~~ documentation not yet written]()
-- [Attributes](docs/attributes.md)
-- [~~File Attributes~~ documentation not yet written]()
-- [~~Streams~~ documentation not yet written]()
+- cmd: The command name, must be `FileWriter_new`.
+- job_id: A unique identifier for the request, a UUID for example.
+- broker: The Kafka broker to use for data.
+- start_time: The start time in milliseconds (UNIX epoch) for data to be written from. Optional, if not supplied then the Kafka message time is used.
```
SkyToGround:
```suggestion
- start_time: The start time in milliseconds (UNIX epoch) for data to be written from. Optional, if not supplied then the timestamp of the Kafka message containing the start command is used.
```
Note: How does this work if the start command is supplied via JSON-file?

documentation/commands.md
```diff
@@ -1,233 +1,243 @@
...
+- stop_time: The stop time in milliseconds (UNIX epoch) for data writing to stop. Optional, if not supplied then file writing continues until a stop command is received
+- service_id: The identifier for the instance of the file-writer that should handle this command. Optional, only needed if multiple file-writers present.
+- abort_on_uninitialised_stream: Whether to abort if the stream cannot be initialised. Optional, default is not to abort but carry on.
```
SkyToGround: What does "cannot be initialised" mean?

documentation/commands.md
```diff
@@ -1,233 +1,243 @@
...
-  "service_id": "[OPTIONAL] the_name_of_the_instance_which_should_interpret_this_command"
-}
-```
+In the example above, a group called my_test_group is defined which in turn has a stream and an attribute defined.
+The stream definition instructs the file-writer that there is data available which should be written to the containing group.
+The parameters for the stream definition are:
 
-#### Command to stop a file being written
+- writer_module: The FlatBuffers schema in the [streaming-data-types](https://github.com/ess-dmsc/streaming-data-types) repository that was used to serialise the data.
+- type: The type of the data. The possible types are defined in the schema declared in the `writer_module`.
```
SkyToGround: We should have some information on the different requirements of `type` for the different writer modules but maybe that is outside the scope of this PR.

documentation/commands.md
```diff
@@ -1,233 +1,243 @@
...
 
-```json
-{
-  "cmd": "FileWriter_stop",
-  "job_id": "job-unique-identifier",
-  "stop_time" : "[OPTIONAL] timestamp (int) in milliseconds",
-  "service_id": "[OPTIONAL] the_name_of_the_instance_which_should_interpret_this_command"
-}
-```
+Note: streams are automatically assigned a NeXus class based on the `writer_module`. [NEEDS TO BE CHECKED]
```
SkyToGround: No need to check, some writer modules automatically assigns NeXus class, some does not. See also [DM-1220](https://jira.esss.lu.se/browse/DM-1220?jql=project%20%3D%2016000%20AND%20labels%20%3D%20%22FW%26FW%22).

documentation/commands.md
matthew-d-jones: :+1: It would be good to elaborate on that. Specifying an inaccessible broker in a stream URL, or a non-existent writer module are a couple of things which can cause this I think. There may be other possibilities.

documentation/commands.md
matthew-d-jones: There may even be multiple tickets about this because I think I created one on github at some stage.

documentation/commands.md
mattclarke: It uses the system clock

documentation/commands.md
mattclarke: Remind me to nominate it for next time.

documentation/commands.md
mattclarke: Agreed on both counts

documentation/commands.md
mattclarke: Basically anything that could go wrong: mistakes in the JSON, missing keys, no broker, multiple HDF issues...

documentation/commands.md
```diff
@@ -241,3 +251,37 @@ Also:
...
+"attributes": {
+  "units": "Kelvin",
+  "error": 0.02
+}
+```
+Where possible the file-writer will try to infer the type of the data, in this example: a float.
+
+For strings and arrays it is necessary to specify the type, for example:
+
+```json
```
SkyToGround: One final question: Do you know if it is possible to (somehow) use key-value pairs as well as arrays? If it is, how do you do that?

documentation/commands.md
```diff
@@ -241,3 +251,37 @@ Also:
...
+"attributes": {
+  "units": "Kelvin",
+  "error": 0.02
+}
+```
+Where possible the file-writer will try to infer the type of the data, in this example: a float.
+
+For strings and arrays it is necessary to specify the type, for example:
```
SkyToGround:
```suggestion
If more complicated types (i.e. arrays and strings)  are required, all attributes for a given dataset or group must be set using a long form version of the attributes.  In this version, each attribute is described using a set of  key-value pairs in an attribute specific dictionary. This dictionary is contained in a list of (possibly) many dictionaries . Example:
```

PR #405, opened 2019-02-12
Dm 1153 classes to enums

mchorazak:
### Issue

*[DM-1153](https://jira.esss.lu.se/secure/RapidBoard.jspa?rapidView=659&projectKey=DM&view=detail&selectedIssue=DM-1153)*

### Description of work

- turned `InitResult`(`HDFWriterModule.h`) into `enum class`.

- deleted WriteResult:
 Earlier when writing was successful an `OK` status was retured.
 On error, one of few variants(some were unused) was returned. All kinds of errors were caught and handled in a single `if(X.isError())`. A `timestamp` and error string could be returned, timestamp was populated, but never used. Error message was used.
I created custom `WriterException` which is thrown in places where any kind of error status was returned before. It's handled in place where errors where previously handled.
On success methods just finish - they are void now.

Many tests had to be updated.

### DIFF
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -39,7 +39,7 @@ ProcessMessageResult Source::process_message(FlatbufferMessage const &Message) {
       }
       return ProcessMessageResult::OK;
     } catch (const HDFWriterModuleRegistry::WriterException &E) {
-      LOG(Sev::Debug, "Failure while writing message: {}", E.what());
+      LOG(Sev::Error, "Failure while writing message: {}", E.what());
       return ProcessMessageResult::ERR;
     }
   }
--- a/src/Source.cpp
+++ b/src/Source.cpp
@@ -31,14 +31,13 @@ ProcessMessageResult Source::process_message(FlatbufferMessage const &Message) {
       return ProcessMessageResult::ERR;
     }
     try {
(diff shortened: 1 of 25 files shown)


src/Source.cpp
```diff
@@ -29,19 +30,18 @@ ProcessMessageResult Source::process_message(FlatbufferMessage const &Message) {
...
+    try {
+      WriterModule->write(Message);
+      _cnt_msg_written += 1;
+      _processed_messages_count += 1;
+      if (HDFFileForSWMR != nullptr) {
+        HDFFileForSWMR->SWMRFlush();
       }
+      return ProcessMessageResult::OK;
+    } catch (const HDFWriterModuleRegistry::WriterException &E) {
+      LOG(Sev::Debug, "Failure while writing message: {}", E.what());
```
matthew-d-jones: I think this should be `Sev::Error`. Data not ending up in the file is v bad.
Looks like that was the behaviour before: https://github.com/ess-dmsc/kafka-to-nexus/pull/405/files#diff-b5e422a9fabe9330754618aaf296930eL45

matthew-d-jones:
Spurious failure on Mac, which passed on the branch build, so going ahead with merge.

PR #407, opened 2019-02-14
Incorrect f142 logging and python dtype in commands

mchorazak:
### Issue

[DM-1149](https://jira.esss.lu.se/browse/DM-1149)
[DM-1180](https://jira.esss.lu.se/browse/DM-1180)

### Description of work

Two small tickets in one pull request:

- clearer error logging for f142
- json commands with `dtype` instead of `type` are accepted and parsed.

Updated one test json command to cover the functionality.

### DIFF
--- a/src/HDFFile.cpp
+++ b/src/HDFFile.cpp
@@ -276,12 +276,9 @@ void HDFFile::writeArrayOfAttributes(hdf5::node::Node const &Node,
             Encoding = hdf5::datatype::CharacterEncoding::ASCII;
           }
         }
-        if (auto AttrType = find<std::string>("type", Attribute)) {
-          DType = AttrType.inner();
-          writeAttrOfSpecifiedType(DType, Node, Name, StringSize, Encoding,
-                                   Values);
-        } else if (auto AttrType = find<std::string>("dtype", Attribute)) {
-          DType = AttrType.inner();
+
+        // get type/dtype
+        if (findType(Attribute, DType)) {
           writeAttrOfSpecifiedType(DType, Node, Name, StringSize, Encoding,
                                    Values);
         } else {
@@ -294,6 +291,18 @@ void HDFFile::writeArrayOfAttributes(hdf5::node::Node const &Node,
       }
     }
   }
+}
+
+bool HDFFile::findType(const nlohmann::basic_json<> Attribute,
+                       std::string &DType) {
+  if (auto AttrType = find<std::string>("type", Attribute)) {
+    DType = AttrType.inner();
+    return true;
+  } else if (auto AttrType = find<std::string>("dtype", Attribute)) {
+    DType = AttrType.inner();
+    return true;
+  } else
+    return false;
 }
 
 void writeAttrStringVariableLength(hdf5::node::Node const &Node,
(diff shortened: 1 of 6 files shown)


src/HDFFile.cpp
```diff
@@ -280,6 +280,10 @@ void HDFFile::writeArrayOfAttributes(hdf5::node::Node const &Node,
           DType = AttrType.inner();
           writeAttrOfSpecifiedType(DType, Node, Name, StringSize, Encoding,
                                    Values);
+        } else if (auto AttrType = find<std::string>("dtype", Attribute)) {
```
mattclarke: Pretty much a duplicate of the code above. I would look at unifying both "finds" out into a separate method or something similar.

src/HDFFile.cpp
```diff
@@ -668,6 +672,9 @@ void HDFFile::writeDataset(hdf5::node::Group const &Parent,
     if (auto DatasetTypeObject =
             find<std::string>("type", DatasetInnerObject)) {
       DataType = DatasetTypeObject.inner();
+    } else if (auto DatasetTypeObject =
+                   find<std::string>("dtype", DatasetInnerObject)) {
+      DataType = DatasetTypeObject.inner();
```
mattclarke: Same method then might be used here.

src/HDFFile.cpp
```diff
@@ -758,6 +765,30 @@ void HDFFile::createHDFStructures(
...
+          }
+        }
+      }
+      if (Type == "stream") {
+        string pathstr;
+        for (auto &x : Path) {
+          pathstr += "/" + x;
+        }
+
+        HDFStreamInfo.push_back(StreamHDFInfo{pathstr, Value->dump()});
```
mattclarke: emplace_back

src/schemas/f142/f142_rw.cpp
```diff
@@ -148,6 +148,10 @@ void HDFWriterModule::parse_config(std::string const &ConfigurationStream,
 
   if (auto TypeNameMaybe = find<std::string>("type", ConfigurationStreamJson)) {
     TypeName = TypeNameMaybe.inner();
+    // handle python style dtype
+  } else if (auto TypeNameMaybe =
+                 find<std::string>("dtype", ConfigurationStreamJson)) {
+    TypeName = TypeNameMaybe.inner();
```
mattclarke: Duplication

src/HDFFile.cpp
```diff
@@ -758,6 +765,30 @@ void HDFFile::createHDFStructures(
...
+    } else if (auto TypeMaybe = find<std::string>("dtype", *Value)) {
+      auto Type = TypeMaybe.inner();
+      if (Type == "group") {
+        if (auto NameMaybe = find<std::string>("name", *Value)) {
+          auto Name = NameMaybe.inner();
+          try {
+            hdf_this = Parent.create_group(Name, LinkCreationPropertyList);
+            Path.push_back(Name);
+          } catch (...) {
+            LOG(Sev::Critical, "failed to create group  Name: {}", Name);
```
mattclarke: Perhaps a discussion for another day, but I think we need to stop catching by ...
It seems unlikely to me that any library we use is going to throw something that doesn't inherit from std::exception and if it does we should put a wrapper around it so it doesn't propagate into our code.

PR #410, opened 2019-02-26
Remove potential divide-by-zero error. 

mattclarke:
### Issue

DM-1373

### Description of work

Checks for one message, if so returns zero for the standard deviation (test already exists).
Also, no need to pass doubles as const ref.


matthew-d-jones:
Connection failure in Mac build in the pr-merge build, but it already passed in the branch build so going ahead with merge.

PR #413, opened 2019-03-13
Slimming down docker image

rerpha:
### Issue

relates to DM-1406

### Description of work

I have bundled steps into layers as more layers in the docker image means more space when building. This approach shouldn't take any longer than usual. The image will rebuild when the conanfile updates, however as this only happens every so often and it already causes the build time to increase I don't think this is an issue.

When merged: create and push new image to docker hub

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,13 @@
 .gitignore
 ansible
 /system-tests/output-files/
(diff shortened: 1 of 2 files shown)


rerpha:
This is probably blocked by DM-1421

PR #417, opened 2019-03-15
Flush only if data available

dominikwerder:
### Issue

- Closes DM-1261

### Description of work

- Only attempt to flush if data is available
- Was not an actual bug, but triggered code analysis lint
- ... and with this update we don't do unnecessary work, too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -231,7 +231,7 @@ node {
     }
   }
 
-  //builders['macOS'] = get_macos_pipeline()
+  builders['macOS'] = get_macos_pipeline()
 
   if ( env.CHANGE_ID ) {
       builders['system tests'] = get_system_tests_pipeline()
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -231,7 +231,7 @@ node {
     }
   }
 
-  builders['macOS'] = get_macos_pipeline()
+  //builders['macOS'] = get_macos_pipeline()
 
   if ( env.CHANGE_ID ) {
       builders['system tests'] = get_system_tests_pipeline()
--- a/src/h5.cpp
+++ b/src/h5.cpp
@@ -436,6 +436,9 @@ append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
 }
 
 template <typename T> AppendResult h5d_chunked_1d<T>::flush_buf() {
+  if (buf_n == 0) {
+    return AppendResult::OK;
+  }
   auto wr = DataSet.append_data_1d(reinterpret_cast<T *>(buf.data()),
                                    buf_n / sizeof(T));
   if (wr.status != AppendResult::OK) {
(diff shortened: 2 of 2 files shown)


src/h5.cpp
```diff
@@ -436,6 +436,9 @@ append_ret h5d_chunked_1d<T>::append_data_1d(T const *data, hsize_t nlen) {
 }
 
 template <typename T> AppendResult h5d_chunked_1d<T>::flush_buf() {
+  if (buf_n == 0) {
+    return AppendResult::OK;
+  }
```
matthew-d-jones: I'd rather avoid having more exit points from the function by wrapping the existing code in `if (buf_n > 0)` instead.

Jenkinsfile
```diff
@@ -231,7 +231,7 @@ node {
     }
   }
 
-  builders['macOS'] = get_macos_pipeline()
+  //builders['macOS'] = get_macos_pipeline()
```
matthew-d-jones: I think it is best to keep this enabled. The mac build usually passes on at least one of the two builds which Jenkins does, and we can go ahead and merge in this case.

src/h5.cpp
dominikwerder: The LLVM coding style favors early exits.

If a function has multiple exit points in nested logic, then I'd also vote to reduce the exit points.

But early-exit on pre-condition test is IMHO fine.

matthew-d-jones: "early exit" means don't execute any other code before exiting.
I'm proposing:
```cpp
template <typename T> AppendResult h5d_chunked_2d<T>::flush_buf() {
  if (buf_n > 0) {
    auto wr = DataSet.append_data_2d(reinterpret_cast<T *>(buf.data()),
                                     buf_n / sizeof(T));
    if (wr.status != AppendResult::OK) {
      return wr.status;
    }
    buf_n = 0;
  }
  return AppendResult::OK;
}
```

dominikwerder: > "early exit" means don't execute any other code before exiting.

I don't understand this comment. Isn't that exactly the case in my proposed version?

```cpp
template <typename T> AppendResult h5d_chunked_1d<T>::flush_buf() {
  if (buf_n == 0) {
    return AppendResult::OK;
  }
```

Nothing gets done before that return.

dominikwerder: For reference:

[LLVM style guide on early exit](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code)

dominikwerder:
Both `branch` and `PR` tests have passed, only OSX hangs in both cases.

src/h5.cpp
matthew-d-jones: LLVM does say "if the body of the 'if' is large". But it isn't worth further disagreement so let's get this merged.
I didn't mean to imply that your code didn't exit early, just that my suggestion exits just as early and avoids multiple exit points, which is why I prefer it.

matthew-d-jones:
It's still executing on OSX, I'll merge as soon as it has passed on one of the jobs.

PR #415, opened 2019-03-14
Add graylog if available.

SkyToGround:
### Issue

DM-1449

### Description of work

As the title says. Awaiting comments during our next meeting on Friday.

### Nominate for Group Code Review

- [ ] Nominate for code review

PR #419, opened 2019-03-19
DM-1168 spdlogging

mchorazak:
### Issue

[DM-1168](https://jira.esss.lu.se/browse/DM-1168)

Closes #415 
Closes #328 

#### Overview

- instead of spdlog and graylog separately, we are now using [spdlog with graylog sink plugin](https://github.com/ess-dmsc/conan-spdlog), credits to @rerpha 
- Instead of using a macro, each class holds a `shared_ptr` to logger instance. 
- Some free functions simply call `spdlog::get("filewriterlogger")` if they are unlikely to be used frequently, otherwise a shared pointer is passed as an argument. 
- Logger always prints to console, optionally can also save logs to file and/or send to graylog.
- In `UnitTests.cpp` there are two loggers set up: 
-- Main `filewriterlogger` displaying messages from filewriter. It is turned off by default.
-- `testlogger` that displays only messages from tests.

#### Graylog/no graylog?
Filewriter works also with plain spdlog package. In this case `#ifdef`s in the code prevent from using graylog functionality.
To avoid accidental changes that would break the code for those not using graylog, a new jenkins job is set up that tests the code against conanfile with basic spdlog package.

- [x] Nominate for code review


src/Source.cpp
```diff
@@ -39,22 +38,27 @@ ProcessMessageResult Source::process_message(FlatbufferMessage const &Message) {
...
+      Logger->error("Failure while writing message: {}", E.what());
       return ProcessMessageResult::ERR;
     }
   }
   return ProcessMessageResult::ERR;
 }
 
 void Source::close_writer_module() {
+  // TODO: should log here using Source::Logger field but it's been std::moved
+  // in CommandHandler::addStreamSourceToWriterModule() and would be NULL here
```
matthew-d-jones: Using `spdlog::get()` is fine in this function as it is executed rarely. This TODO comment can be removed.

src/h5.cpp
```diff
@@ -14,23 +14,21 @@ void h5d::init_basics() {
...
   sext = DSPTgt.current_dimensions();
   ShapeMax = DSPTgt.maximum_dimensions();
-  if (log_level >= 9) {
-    for (int i1 = 0; i1 < ndims; ++i1) {
-      LOG(Sev::Debug, "{:20} i: {}  sext: {:21}  ShapeMax: {:21}", Name, i1,
-          sext.at(i1), ShapeMax.at(i1));
+  for (int i1 = 0; i1 < ndims; ++i1) {
+    Logger->trace("{:20} i: {}  sext: {:21}  ShapeMax: {:21}", Name, i1,
+                  sext.at(i1), ShapeMax.at(i1));
+    try {
```
matthew-d-jones: Something has gone wrong here, the `try`-`catch` should not be inside the `for` loop. (`for` loop should only contain the `Logger->trace()` call)

src/h5.cpp
```diff
@@ -181,19 +176,17 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
...
 
-  if (log_level >= 9) {
-    LOG(Sev::Debug, "try to get the dsp dims:");
-    auto sext = DSPTgt.current_dimensions();
-    auto smax = DSPTgt.maximum_dimensions();
-    for (int i1 = 0; i1 < ndims; ++i1) {
-      LOG(Sev::Debug, "dimensions: {:20} {}: {:21} {:21}", Name, i1,
-          sext.at(i1), smax.at(i1));
-    }
+  Logger->trace("try to get the dsp dims:");
```
matthew-d-jones: Can you keep the log level check? This function is called a lot so we should be careful about any potential performance hits.

src/h5.cpp
```diff
@@ -245,15 +236,12 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
...
-    if (log_level >= 7) {
-      auto dsp = hdf5::dataspace::Simple(Dataset.dataspace());
-      auto sext = dsp.current_dimensions();
-      auto smax = dsp.current_dimensions();
-      for (int i1 = 0; i1 < ndims; ++i1) {
-        LOG(Sev::Debug, "dimensions {}: {:12} {:12}", i1, sext.at(i1),
-            smax.at(i1));
-      }
+    Logger->trace("write failed  ds_name: {}", ds_name);
+    auto dsp = hdf5::dataspace::Simple(Dataset.dataspace());
```
matthew-d-jones: Keep the log level check again here if possible

src/logger.cpp
```diff
@@ -1,158 +1,39 @@
...
-#include <cstdarg>
-#include <cstdio>
-#include <cstdlib>
-#include <cstring>
-#include <memory>
 #include <string>
-#include <thread>
 #ifdef HAVE_GRAYLOG_LOGGER
 #include <graylog_logger/GraylogInterface.hpp>
 #include <graylog_logger/Log.hpp>
```
matthew-d-jones: I suspect the `graylog_logger` headers don't need to be included here anymore.

src/logger.h
```diff
@@ -1,59 +1,14 @@
...
-                fmt::format("ERROR in format: {}: {}", e.what(), fmt));
-  }
-}
-
-void use_log_file(std::string fname);
-
-void log_kafka_gelf_start(std::string const &Address, std::string TopicName);
-
-void fwd_graylog_logger_enable(std::string const &Address);
+// clang-format off
```
matthew-d-jones: Why does clang-format need to be off for these includes?

src/schemas/f142/f142_rw.cpp
```diff
@@ -142,11 +144,13 @@ void HDFWriterModule::parse_config(std::string const &ConfigurationStream,
           find<std::string>("source", ConfigurationStreamJson)) {
     SourceName = SourceNameMaybe.inner();
   } else {
-    LOG(Sev::Error, "Key \"source\" is not specified in json command");
+    Logger->error("Key \"source\" is not specified in json command");
     return;
   }
 
-  if (!findType(ConfigurationStreamJson, TypeName)) {
+  if (auto TypeNameMaybe = find<std::string>("type", ConfigurationStreamJson)) {
```
matthew-d-jones: This looks like a mistake again, the `dtype` support change is being reverted in this branch for some reason.

src/logger.h
mchorazak: `spdlog/spdlog.h` needs to be included before any other spdlog header. Clang-format reverts the order.

src/logger.cpp
mchorazak: Not including `graylog_logger` headers breaks the build. Despite using spdlog-graylog package you still need to explicitly include libraries both in cmake and here.

src/logger.h
matthew-d-jones: Ah, can you explain in a comment then please.

matthew-d-jones:
@SkyToGround 
:man_facepalming: I didn't know/remember that we were logging those details.
We can do something like:
```cpp
#define LOG_TRACE(logger, ...) if (logger->should_log(level::trace)){logger->trace("{}::{}()#{}: ", __FILE__ , __FUNCTION__, __LINE__, fmt::format(__VA_ARGS__));}
```
Or given that we don't require different log levels for different parts of the application, we could go back to a global logger instead of having a local logger as a macro argument.

Either way doesn't present a problem with spdlog thankfully. We can discuss this at the review on Monday.

matthew-d-jones:
The log format should be set so that message look the same as before the change. This is done something like:
```cpp
spdlog::set_pattern("[%H:%M:%S %z] [%n] [%^---%L---%$] [thread %t] %v");
```
(though that is not the correct pattern)

matthew-d-jones:
We should also check how frequently the file is flushed when logging to file. If necessary this can be controlled explicitly:
```cpp
spdlog::flush_every(std::chrono::seconds(5));
```

mchorazak:
Everything is now done in a simpler way, less code duplication etc.
I failed to find out why doesn't graylogger display `trace` and `debug` levels. Higher levels work just fine with `info` being reported by grayloger as `notice` as could be expected because of this map in `graylog_sink.h`:
```
std::map<int, Log::Severity> LogLevels{
      {SPDLOG_LEVEL_TRACE, Log::Severity::Debug},
      {SPDLOG_LEVEL_DEBUG, Log::Severity::Info},
      {SPDLOG_LEVEL_INFO, Log::Severity::Notice},
      {SPDLOG_LEVEL_WARN, Log::Severity::Warning},
      {SPDLOG_LEVEL_ERROR, Log::Severity::Error},
      {SPDLOG_LEVEL_CRITICAL, Log::Severity::Critical},
      {SPDLOG_LEVEL_OFF, Log::Severity::Alert}};```

I will look into it Monday morning.

**Comment by SkyToGround** [comment]
@mchorazak I am pretty sure I know why Graylog does not print some log messages. It is probably because you do not set (its) log level anywhere. Ask me about it on Monday if you need help.

File: src/logger.h [hunk 8]
```diff
@@ -1,59 +1,20 @@
...
+// spdlog.h has to be included before sink headers. Clang-format reverts the
+// order so it needs to be turned off.
+// clang-format off
+#include <spdlog/spdlog.h>
+#include <spdlog/sinks/basic_file_sink.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+// clang-format on
 
-#ifdef _MSC_VER
+typedef std::shared_ptr<spdlog::logger> SharedLogger;
```
SkyToGround:
```suggestion
using SharedLogger = std::shared_ptr<spdlog::logger>;
```

src/logger.h
```diff
@@ -1,59 +1,20 @@
...
 
-#define LOG(level, fmt, ...)                                                   \
-  { dwlog(level, fmt, __FILE__, __LINE__, __FUNCSIG__, __VA_ARGS__); }
+std::shared_ptr<spdlog::logger> getLogger();
```
SkyToGround:
```suggestion
SharedLogger getLogger();
```

src/CLIOptions.cpp
```diff
@@ -103,17 +132,22 @@ void setCLIOptions(CLI::App &App, MainOpt &MainOptions) {
       ->required();
   addUriOption(App, "--status-uri", MainOptions.KafkaStatusURI,
                MainOptions.ReportStatus,
-               "<host[:port][/topic]> Kafka broker/topic to publish status "
+               "<//host[:port][/topic]> Kafka broker/topic to publish status "
```
SkyToGround:
```suggestion
               "<host[:port][/topic]> Kafka broker/topic to publish status "
```
To avoid reverting recent changes.

src/DemuxTopic.cpp
```diff
@@ -8,7 +8,7 @@ namespace FileWriter {
 
 DemuxTopic::DemuxTopic(std::string TopicName) : Topic(std::move(TopicName)) {}
 
-DemuxTopic::~DemuxTopic() { LOG(Sev::Debug, "dtor"); }
+DemuxTopic::~DemuxTopic() { Logger->trace("DemuxTopic destructor"); }
```
SkyToGround: 👍

src/h5.cpp
```diff
@@ -181,18 +180,18 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
...
+  if (Logger->should_log(spdlog::level::trace)) {
+    Logger->trace("try to get the dsp dims:");
     auto sext = DSPTgt.current_dimensions();
     auto smax = DSPTgt.maximum_dimensions();
     for (int i1 = 0; i1 < ndims; ++i1) {
-      LOG(Sev::Debug, "dimensions: {:20} {}: {:21} {:21}", Name, i1,
-          sext.at(i1), smax.at(i1));
+      Logger->trace("dimensions: {:20} {}: {:21} {:21}", Name, i1, sext.at(i1),
+                    smax.at(i1));
     }
```
SkyToGround:
```suggestion
```

src/h5.cpp
```diff
@@ -181,18 +180,18 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
...
-        std::chrono::duration_cast<MS>(t2 - t1).count(),
-        std::chrono::duration_cast<MS>(t3 - t2).count());
+    Logger->trace("h5d::append_data_1d set_extent: {} + {}",
+                  std::chrono::duration_cast<MS>(t2 - t1).count(),
+                  std::chrono::duration_cast<MS>(t3 - t2).count());
   }
 
-  if (log_level >= 9) {
-    LOG(Sev::Debug, "try to get the dsp dims:");
+  if (Logger->should_log(spdlog::level::trace)) {
```
SkyToGround:
```suggestion
```

src/h5.cpp
```diff
@@ -245,14 +242,14 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
   try {
     Dataset.write(*data, Type, DSPMem, DSPTgt, PLTransfer);
   } catch (...) {
-    LOG(Sev::Error, "write failed  ds_name: {}", ds_name);
-    if (log_level >= 7) {
+    if (Logger->should_log(spdlog::level::trace)) {
+      Logger->trace("write failed  ds_name: {}", ds_name);
```
SkyToGround:
```suggestion
```

src/h5.cpp
```diff
@@ -245,14 +242,14 @@ append_ret h5d::append_data_1d(T const *data, hsize_t nlen) {
   try {
     Dataset.write(*data, Type, DSPMem, DSPTgt, PLTransfer);
   } catch (...) {
-    LOG(Sev::Error, "write failed  ds_name: {}", ds_name);
-    if (log_level >= 7) {
+    if (Logger->should_log(spdlog::level::trace)) {
```
SkyToGround:
```suggestion
    Logger->error("write failed  ds_name: {}", ds_name);
    if (Logger->should_log(spdlog::level::trace)) {
```

src/h5.cpp
matthew-d-jones: I think this `if` should be there, it is preventing getting dimensions and executing a loop if the log level is set above `trace`, which would usually be the case.

SkyToGround: I think that this is a false optimisation. Better to remove useless code.

src/schemas/NDAr/AreaDetectorWriter.h
```diff
@@ -70,5 +70,8 @@ class AreaDetectorWriter : public FileWriterBase {
   int CueCounter{0};
   NeXusDataset::CueIndex CueTimestampIndex;
   NeXusDataset::CueTimestampZero CueTimestamp;
+
+private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/ev42/ev42_rw.h
```diff
@@ -11,6 +11,9 @@ class FlatbufferReader : public FileWriter::FlatbufferReader {
   bool verify(FlatbufferMessage const &Message) const override;
   std::string source_name(FlatbufferMessage const &Message) const override;
   uint64_t timestamp(FlatbufferMessage const &Message) const override;
+
+private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/ev42/ev42_rw.h
```diff
@@ -38,6 +41,9 @@ class HDFWriterModule : public FileWriter::HDFWriterModule {
   uint64_t ts_max = 0;
   size_t buffer_size = 0;
   size_t buffer_packet_max = 0;
+
+private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/ev42/ev42_synth.h
```diff
@@ -28,6 +29,9 @@ class synth {
   FlatBufferWrapper next(uint32_t size);
   std::unique_ptr<SynthImpl> impl;
   std::string Name;
+
+private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/f142/WriterArray.h
```diff
@@ -21,6 +21,7 @@ template <typename DT, typename FV> class WriterArray : public WriterTypedBase {
   uptr<h5::h5d_chunked_2d<DT>> ChunkedDataset;
   Value FlatbuffersValueTypeId = Value::NONE;
   size_t ChunkSize = 64 * 1024;
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/f142/WriterScalar.h
```diff
@@ -21,6 +21,7 @@ class WriterScalar : public WriterTypedBase {
   uptr<h5::h5d_chunked_1d<DT>> ChunkedDataset;
   Value FlatbuffersValueTypeId = Value::NONE;
   size_t ChunkSize = 64 * 1024;
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/f142/WriterScalarString.h
```diff
@@ -15,6 +15,7 @@ class WriterScalarString : public WriterTypedBase {
   h5::append_ret write(FBUF const *fbuf) override;
   h5::Chunked1DString::ptr ChunkedDataset;
   Value FlatbuffersValueTypeId = Value::String;
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/f142/f142_rw.h
```diff
@@ -110,6 +110,7 @@ class HDFWriterModule final : public FileWriter::HDFWriterModule {
                                                        ErrorLogMinInterval};
 
 private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/h5.cpp
SkyToGround: @matthew-d-jones By the way, committing those changes was actually a mistake by me. So if you have a strong opinion about this, I can revert it.

src/schemas/hs00/Writer.h
```diff
@@ -27,6 +27,9 @@ class Writer : public FileWriter::HDFWriterModule {
   hsize_t ChunkBytes = 1 << 21;
   bool DoFlushEachWrite = true;
   uint64_t TotalWrittenBytes = 0;
+
+private:
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/hs00/WriterTyped.h
```diff
@@ -80,11 +80,12 @@ class WriterTyped : public WriterUntyped {
 
   uint64_t LargestTimestampSeen = 0;
   size_t MaxNumberHistoric = 4;
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/schemas/senv/FastSampleEnvironmentWriter.h
```diff
@@ -53,5 +53,6 @@ class FastSampleEnvironmentWriter : public FileWriterBase {
   NeXusDataset::Time Timestamp;
   NeXusDataset::CueIndex CueTimestampIndex;
   NeXusDataset::CueTimestampZero CueTimestamp;
+  std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
  SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/tests/HDFFile.cpp
```diff
@@ -400,11 +403,14 @@ class T_CommandHandler : public testing::Test {
...
         if (msgs.back().size() < 8) {
-          LOG(Sev::Error, "error");
+          Logger->error("error");
           exit(1);
         }
       }
     }
+
+  private:
+    std::shared_ptr<spdlog::logger> Logger = spdlog::get("testlogger");
```
SkyToGround:
```suggestion
    SharedLogger Logger = spdlog::get("testlogger");
```

src/tests/HDFFile.cpp
```diff
@@ -420,6 +426,7 @@ class T_CommandHandler : public testing::Test {
   }
 
   static void data_ev42() {
+    std::shared_ptr<spdlog::logger> Logger = spdlog::get("filewriterlogger");
```
SkyToGround:
```suggestion
    SharedLogger Logger = spdlog::get("filewriterlogger");
```

src/tests/HDFFile.cpp
```diff
@@ -791,9 +797,13 @@ class T_CommandHandler : public testing::Test {
             fb.builder->GetSize()));
       }
     }
+
+  private:
+    std::shared_ptr<spdlog::logger> Logger = spdlog::get("testlogger");
```
SkyToGround:
```suggestion
   SharedLogger Logger = spdlog::get("testlogger");
```

src/tests/HDFFile.cpp
```diff
@@ -791,9 +797,13 @@ class T_CommandHandler : public testing::Test {
...
   };
 
   static void data_f142() {
+    std::shared_ptr<spdlog::logger> Logger = spdlog::get("testlogger");
```
SkyToGround:
```suggestion
    SharedLogger Logger = spdlog::get("testlogger");
```

src/h5.cpp
SkyToGround: The suggested change does not look like it should here. The thing is that  the _write failed_ message was changed from being an error to being a a trace message here and it was moved from being outside the if-statement to being inside it. I moved it back outside and changed the severity back into _error_.

PR #420, opened 2019-03-22
Minor refactoring

matthew-d-jones:
### Description of work

Small refactoring changes for things I came across when looking into another ticket:
- Remove superfluous includes
- Move includes to implementation file from header
- Change produce method to take a string instead of having reinterpret and c-style casts in many places
- Improve erroneous or waffling comments
- Don't use `ASSERT_DOUBLE_EQ` for integer types

### DIFF
--- a/src/EventLogger.h
+++ b/src/EventLogger.h
@@ -63,7 +63,6 @@ void logEvent(std::shared_ptr<ProducerType> Producer, StatusCode Code,
   Event["message"] = Message;
 
   std::string EventMessage = Event.dump();
-  Producer->produce(reinterpret_cast<unsigned char *>(&EventMessage[0]),
-                    EventMessage.size());
+  Producer->produce(EventMessage);
 }
 }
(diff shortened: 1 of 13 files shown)


src/EventLogger.h
```diff
@@ -63,7 +63,6 @@ void logEvent(std::shared_ptr<ProducerType> Producer, StatusCode Code,
   Event["message"] = Message;
 
   std::string EventMessage = Event.dump();
-  Producer->produce(reinterpret_cast<unsigned char *>(&EventMessage[0]),
-                    EventMessage.size());
+  Producer->produce(EventMessage);
```
SkyToGround: 👍

src/Master.cpp
```diff
@@ -152,7 +152,7 @@ void Master::statistics() {
     Status["files"][FilewriterTaskID] = FilewriterTaskStatus;
   }
   auto Buffer = Status.dump();
```
rerpha: might be best to use std::string here and be explicit,  like the other `dump()` call in logger.cpp:125

matthew-d-jones: Thanks, made the change in another branch.

PR #425, opened 2019-04-25
Update cppcheck and minor fixes

mattclarke:
### Issue

N/A

### Description of work

Updated CppCheck to be the same as the Forwarder and fixed some of the warnings.


CMakeLists.txt
```diff
@@ -35,6 +35,12 @@ elseif(NOT ${CONAN} MATCHES "DISABLE")
   MESSAGE(FATAL_ERROR "Unrecognised option for CONAN, use AUTO, MANUAL or DISABLE")
 endif()
 
+# Yes, there exists also FindGTest.
+# Yes, we do want to use this one instead:
+if(NOT GOOGLETEST_DISABLE)
```
matthew-d-jones: Should this use the `BUILD_TESTS` parameter you added to the readme?

src/HDFFile.cpp
```diff
@@ -242,7 +242,8 @@ void writeAttributes(hdf5::node::Node const &Node, nlohmann::json const *Value,
 /// \param JsonValue    json value array of attribute objects.
 /// \param Logger Pointer to spdlog instance to be used for logging.
 void writeArrayOfAttributes(hdf5::node::Node const &Node,
-                            const nlohmann::json &Values, SharedLogger Logger) {
+                            const nlohmann::json &Values,
+                            SharedLogger const &Logger) {
```
matthew-d-jones: East const, west const, east const -.-
Maybe we should standardise?

mattclarke: East const for me

mattclarke: That also seems to be most prevalent.

matthew-d-jones: Ok, fine with me.
I'll do my best to break my west const habit... I think Jack also picked up west const from me. Jonas uses east const though. Will try to remember to bring it up in a meeting.

PR #427, opened 2019-04-29
Dm 1519 do not exit if broker not available

mchorazak:
### Issue

DM-1519

### Description of work

If broker is unavailable, filewriter tries to establish connection every ~500 ms. Every ~10s an error is logged.


src/KafkaW/Consumer.cpp
```diff
@@ -129,13 +130,29 @@ bool Consumer::topicPresent(const std::string &TopicName) {
 
 void Consumer::updateMetadata() {
```
matthew-d-jones: Now this function is doing something more complicated it would be nice to have a doxygen comment to explain.

src/KafkaW/Consumer.cpp
```diff
@@ -129,13 +130,29 @@ bool Consumer::topicPresent(const std::string &TopicName) {
...
-      true, nullptr, &MetadataPtr, ConsumerBrokerSettings.MetadataTimeoutMS);
-  if (ErrorCode == RdKafka::ERR__TRANSPORT)
-    throw MetadataException("Cannot contact broker");
-  else if (ErrorCode != RdKafka::ERR_NO_ERROR) {
-    throw MetadataException(
-        "Consumer::updateMetadata() - error while retrieving metadata.");
+  bool Connected = false;
+  short LoopCounter = 0;
+  while (!Connected) {
+    auto ErrorCode = KafkaConsumer->metadata(
```
matthew-d-jones: The logic for the metadata query itself (basically what the method did before your changes) could be refactored to a separate method, which returns the `Connected` bool. I think that would improve readability as it would separate the query logic from the retry logic.

matthew-d-jones:
Unit tests were running for >14 hours. There must be a call to `updateMetadata` in the tests.

src/StreamMaster.h
```diff
@@ -28,7 +28,6 @@ namespace FileWriter {
 template <typename Streamer> class StreamMaster {
   using StreamerStatus = Status::StreamerStatus;
   using StreamMasterError = Status::StreamMasterError;
-  friend class CommandHandler;
```
SkyToGround: 👍 Extra points for the funny commit message.

mattclarke: 👍

src/KafkaW/ConsumerFactory.cpp
```diff
@@ -1,15 +1,27 @@
 #include "ConsumerFactory.h"
+#include "helper.h"
 
 namespace KafkaW {
 
-std::unique_ptr<Consumer> createConsumer(const BrokerSettings &Settings) {
-  std::string ErrorString;
+std::unique_ptr<Consumer> createConsumer(const BrokerSettings &Settings,
```
SkyToGround: It took me longer than I wanted to wrap my head around what is happening here but after having done so, I determined that there is a design problem that we should fix seeing as we are touching the code anyways.

The `BrokerSettings` class should not be responsible for applying any configuration to a `RdKafka::Conf` object. This should either be done in this function (too ugly), by the `RdKafka::Conf` object itself (too hard) or by a free function (preferred).

SkyToGround: Oh, and the function that implements this functionality should **not** be called `apply`.

src/CommandHandler.cpp
```diff
@@ -320,7 +317,7 @@ void CommandHandler::handleNew(std::string const &Command,
       s->report(std::chrono::milliseconds{Config.StatusMasterIntervalMS});
     }
     if (Config.topic_write_duration.count() != 0) {
-      s->TopicWriteDuration = Config.topic_write_duration;
+      s->setTopicWriteDuration(Config.topic_write_duration);
```
SkyToGround: 👍

src/KafkaW/Consumer.cpp
```diff
@@ -127,17 +133,36 @@ bool Consumer::topicPresent(const std::string &TopicName) {
   return true;
 }
 
+/// Updates the RdKafka Metadata pointer. If broker is unavailable, keeps trying
+/// to connect every 500ms, logging messages every few seconds.
 void Consumer::updateMetadata() {
+  short LoopCounter = 0;
+  while (!metadataCall()) {
```
SkyToGround: Although this function is very simple, I still think that we should have one or two unit tests for it. It should be easy (I think) to write a simple test that mocks the call to `metadataCall()`.

src/KafkaW/Consumer.cpp
```diff
@@ -127,17 +133,36 @@ bool Consumer::topicPresent(const std::string &TopicName) {
...
+    }
+    LoopCounter++;
+  }
+  Logger->info("Connection with broker established.");
+}
+
+bool Consumer::metadataCall() {
   RdKafka::Metadata *MetadataPtr = nullptr;
   auto ErrorCode = KafkaConsumer->metadata(
       true, nullptr, &MetadataPtr, ConsumerBrokerSettings.MetadataTimeoutMS);
```
SkyToGround: I have had problems before caused by metadata calls taking too long and timing out before getting a response. Even with a timeout of 1 second. Solving this problem in a satisfactory way is hard as the call to `metadata()` is a blocking call. A `maybe` acceptable work around for now could be to execute that call in a separate thread (with a timeout of let's say 10 seconds) so as to not block the current thread.

As a bare minimum, you should add the following Doxygen comment `/// /todo Re-design function to execute blocking call in separate thread.`.

src/KafkaW/Consumer.cpp
```diff
@@ -127,17 +133,36 @@ bool Consumer::topicPresent(const std::string &TopicName) {
...
-  if (ErrorCode == RdKafka::ERR__TRANSPORT)
-    throw MetadataException("Cannot contact broker");
-  else if (ErrorCode != RdKafka::ERR_NO_ERROR) {
+  switch (ErrorCode) {
+  case RdKafka::ERR_NO_ERROR:
+    KafkaMetadata = std::shared_ptr<RdKafka::Metadata>(MetadataPtr);
+    return true;
+  case RdKafka::ERR__TRANSPORT:
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
SkyToGround: Can you tell me the reasoning for sleeping here?

src/KafkaW/Consumer.cpp
SkyToGround: Reading the code a bit further, it looks like you actually are executing the code in a separate thread already, correct? Regardless, the code still needs some changes to nicely deal with longer timeouts.

src/KafkaW/Consumer.cpp
mchorazak: I wanted to pause the thread only if this particular error occurs. If everything goes right or if another error happens there is no delay.

src/KafkaW/Consumer.cpp
mchorazak: Yes, each consumer is async launched as a separate thread.
We recently also discovered metadata call timing out on kafkacow running on windows. We're not sure what causes it. We'll see if that's related.
We will discuss it with @matthew-d-jones .

src/KafkaW/Consumer.cpp
SkyToGround: @mchorazak Can you be a bit more specific as to why? :-)

matthew-d-jones: The idea was to avoid spamming metadata requests if it fails, but it looks like `RdKafka::ERR__TRANSPORT` only occurs if the timeout is hit anyway.

src/tests/HDFFileTests.cpp
```diff
@@ -99,13 +98,12 @@ class T_CommandHandler : public testing::Test {
...
-    MainOpt main_opt;
-    MockMasterI MMaster;
+    std::string Filename = Command["file_attributes"]["file_name"];
+    std::remove(Filename.c_str());
+    MainOpt MainOptions;
 
-    FileWriter::CommandHandler ch(main_opt, &MMaster);
-    ch.tryToHandle(CommandString);
+    FileWriter::CommandHandler CommandHandlerToTest(MainOptions, nullptr);
+    CommandHandlerToTest.tryToHandle(CommandString);
```
SkyToGround: 👍

src/KafkaW/Producer.cpp
```diff
@@ -26,7 +27,7 @@ Producer::~Producer() {
 
 void Producer::setConf(std::string &ErrorString) {
   try {
-    ProducerBrokerSettings.apply(Conf.get());
+      configureKafka(Conf.get(), ProducerBrokerSettings);
```
SkyToGround: 👍 +1

src/KafkaW/Consumer.cpp
SkyToGround: I have created a JIRA ticket for this issue: https://jira.esss.lu.se/browse/DM-1599
I think this temporarily resolves the issue for but I also think we need to fix the problem as soon as possible.

SkyToGround:
I think this is ready to be merged. We should try to get around to fixing the metadata timeout issue as soon as possible though. Any other comments @matthew-d-jones and @mattclarke ?

matthew-d-jones:
@SkyToGround I'll merge this as you are happy with it now. As Michal mentioned we have also come across an issue with metadata queries now, although we only saw the problem on Windows. It was repeatable thgouh, so it might give us a way to investigate.

PR #439, opened 2019-06-18
Dm 1530 nicos cache module

michele-brambilla:
### Issue

DM-1530

### Description of work

Add a filewriter module that allows to write the ns10 schema. This allows to store in the NeXus file values for non-EPICS devices that live in NICOS (and store data in the KafkaCache).


src/schemas/ns10/NicosCacheWriter.h
```diff
@@ -0,0 +1,72 @@
...
+/// \brief Writing module for the NICOS cache values.
+
+#pragma once
+#include "../../FlatbufferReader.h"
+#include "../../HDFWriterModule.h"
+#include "NeXusDataset.h"
+
+#include <iostream>
+
+namespace NicosCacheWriter {
```
mattclarke: The others have a namespace like:
```
namespace FileWriter {
namespace Schemas {
namespace hs00 {
```
I think this should match that convention

src/schemas/ns10/NicosCacheWriter.h
```diff
@@ -0,0 +1,72 @@
...
+  std::string
+  source_name(FileWriter::FlatbufferMessage const &Message) const override;
+
+  uint64_t
+  timestamp(FileWriter::FlatbufferMessage const &Message) const override;
+
+private:
+};
+
+class CacheWriter : public FileWriter::HDFWriterModule {
```
mattclarke: Split out this class into another file `Writer.h' so the header files and implementation files match up. E.g.
Reader.h and Reader.cpp
Writer.h and Writer.cpp

src/schemas/ns10/Reader.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace NicosCacheWriter {
+
+// Instantiates a ReaderClass used for extracting source names, timestamps and
+// verifying a flatbuffers.
+static FileWriter::FlatbufferReaderRegistry::Registrar<CacheReader>
+    RegisterReader("ns10");
+}
+
+namespace NicosCacheWriter {
```
mattclarke: Why is the namespace declared twice?

src/schemas/ns10/NicosCacheWriter.h
michele-brambilla: The reader consists only of a few lines. What about I merge the two?

mattclarke: I would prefer them separate like was done for hs00.

PR #444, opened 2019-06-20
Fixed some simple cppcheck fixes

mattclarke:
### Issue
N/A

### Description of work

Removing unused vars, suppression etc.


src/tests/EventHistogramWriterTests.cpp
```diff
@@ -300,10 +299,9 @@ createTestMessage(size_t HistogramID, size_t PacketID,
 
   flatbuffers::Offset<void> DataValue;
   {
-    size_t TotalElements = 1;
-    for (auto x : ThisLengths) {
-      TotalElements *= x;
-    }
+    size_t TotalElements = std::accumulate(
+        ThisLengths.begin(), ThisLengths.end(), 1, std::multiplies<>());
```
matthew-d-jones: `std::accumulate` is a pain. I think `1` is treated as an `int` here, then the result of std::accumulate is cast to `size_t`. `1` should be `size_t` to avoid possibility of bad behaviour for really big vectors.

src/tests/EventHistogramWriterTests.cpp
```diff
@@ -329,10 +327,9 @@ createTestMessage(size_t HistogramID, size_t PacketID,
 
   flatbuffers::Offset<void> ErrorValue;
   {
-    size_t TotalElements = 1;
-    for (auto x : ThisLengths) {
-      TotalElements *= x;
-    }
+    size_t TotalElements = std::accumulate(
+        ThisLengths.begin(), ThisLengths.end(), 1, std::multiplies<>());
```
matthew-d-jones: Same as above.

PR #446, opened 2019-06-27
Use hash instead of source name to find FB message sink

SkyToGround:
### Issue

* Closes DM-1129

### Description of work

* Replaced the use of the source name to identify a message sink with a hash of the Flatbuffer ID plus source name.
* Fixed relevant unit tests.
* Added unit test.
* Changed GTest to be linked statically to work with CLion on MacOSX.
* Some minor clang-format related changes/improvements.

It is not strictly necessary to use a hash, just using the strings would be good enough in this case. The performance benefit of using a hash are minimal to non-existent. However, I would argue that a hash is a slightly cleaner design that simplifies (potential) future re-designs.


src/KafkaW/ConfigureKafka.cpp
```diff
@@ -17,4 +17,4 @@ void configureKafka(RdKafka::Conf *RdKafkaConfiguration,
     }
   }
 }
-}
\ No newline at end of file
+} // namespace KafkaW
```
SkyToGround:
```suggestion
} // namespace KafkaW

```

src/DemuxTopic.cpp
```diff
@@ -23,19 +23,19 @@ ProcessMessageResult
...
   try {
-    auto &CurrentSource = TopicSources.at(Message.getSourceName());
+    auto &CurrentSource = TopicSources.at(Message.getSourceHash());
     auto ProcessingResult = CurrentSource.process_message(Message);
     ++messages_processed;
     return ProcessingResult;
   } catch (std::out_of_range &e) {
-    Logger->trace("Source with name \"{}\" is not in list.",
-                  Message.getSourceName());
+    Logger->trace("Source with name \"{}\" and ID \"{}\" is not in list.",
```
matthew-d-jones: Braces don't need to be escaped.

src/Streamer.cpp
```diff
@@ -145,10 +145,9 @@ FileWriter::Streamer::pollAndProcess(FileWriter::DemuxTopic &MessageProcessor) {
     return ProcessMessageResult::ERR;
   }
 
-  if (std::find(Sources.begin(), Sources.end(), Message->getSourceName()) ==
-      Sources.end()) {
-    Logger->warn("Message from topic \"{}\" has an unknown source name "
-                 "(\"{}\"), ignoring.",
+  if (Sources.find(Message->getSourceHash()) == Sources.end()) {
+    Logger->warn("Message from topic \"{}\" with the source name \"{}\" is "
```
matthew-d-jones: Braces don't need to be escaped.

src/DemuxTopic.cpp
SkyToGround: I don't understand, I am not escaping braces here. I am escaping the quotation marks.

src/Streamer.cpp
SkyToGround: Ditto.

src/DemuxTopic.cpp
matthew-d-jones: Oops, somehow missed that ^^

SkyToGround: np

SkyToGround:
I will merge when Jenkins is done then.

PR #465, opened 2019-08-29
Write ADC pulse debug data with ev42 writer module

matthew-d-jones:
### Issue

DM-1276

### Description of work

If `ev42` writer module is configured to write ADC pulse debug data ([see updated documentation](https://github.com/ess-dmsc/kafka-to-nexus/pull/465/files#diff-db0fc51e3050fccd9becf710bf5e2980R26-R30)) and it is present in messages received, then it will do so. ADC pulse data does not have to be present in every event message. There is a configuration option (defaulting to `false`) to chose whether to write this data for two reasons:
- the data volume may be large so you may not always want to write it to file even if it is present in stream.
- I wanted to avoid there always being empty datasets created for it even in the case that ADC pulse data are not being produced, which will presumably always be the case after commissioning an instrument is complete.

The commits from #463 (adding unit tests for `ev42` writer module) will be shown below until that PR is merged.


src/schemas/ev42/ev42_rw.cpp
```diff
@@ -117,26 +112,65 @@ void HDFWriterModule::parse_config(std::string const &ConfigurationStream,
...
+  }
+}
+
+void HDFWriterModule::createAdcGroupAndDatasets(hdf5::node::Group &HDFGroup) {
+  auto AdcGroup = HDFGroup.create_group(AdcGroupName);
+
+  NeXusDataset::Amplitude(        // NOLINT(bugprone-unused-raii)
+      AdcGroup,                   // NOLINT(bugprone-unused-raii)
+      NeXusDataset::Mode::Create, // NOLINT(bugprone-unused-raii)
+      chunk_bytes);               // NOLINT(bugprone-unused-raii)
```
SkyToGround: Chunk size is in this case the number of elements, not number of bytes. Due to the current state of the ev42 writer module, I suggest that we leave the implementation as is but write a comment (and maybe a JIRA ticket) to clarify this fact and that we intend to fix it in the future.

src/schemas/ev42/ev42_rw.cpp
```diff
@@ -117,26 +112,65 @@ void HDFWriterModule::parse_config(std::string const &ConfigurationStream,
...
+  try {
+    RecordAdcPulseDebugData =
+        ConfigurationStreamJson["adc_pulse_debug"].get<bool>();
+    Logger->trace("adc_pulse_debug: {}", RecordAdcPulseDebugData);
+  } catch (...) { /* it's ok if not found */
+  }
+}
+
+void HDFWriterModule::createAdcGroupAndDatasets(hdf5::node::Group &HDFGroup) {
+  auto AdcGroup = HDFGroup.create_group(AdcGroupName);
```
SkyToGround: When discussing the layout of the debug data with Tobias, he suggested that it be put next to the other datasets (i.e. in the `HDFGroup`-group). I do not have a strong opinion on the matter however so I suggest that we leave @zjttoefs until Friday afternoon to comment. We can easily change things later.

src/schemas/ev42/ev42_rw.cpp
```diff
@@ -117,26 +112,65 @@ void HDFWriterModule::parse_config(std::string const &ConfigurationStream,
...
+      chunk_bytes);               // NOLINT(bugprone-unused-raii)
 }
 
 HDFWriterModule::InitResult
 HDFWriterModule::init_hdf(hdf5::node::Group &HDFGroup,
                           std::string const &HDFAttributes) {
 
   try {
-    this->ds_event_time_offset = h5::h5d_chunked_1d<uint32_t>::create(
+    ds_event_time_offset = h5::h5d_chunked_1d<uint32_t>::create(
```
SkyToGround: 👍

src/schemas/ev42/ev42_rw.cpp
```diff
@@ -200,31 +242,83 @@ HDFWriterModule::reopen(hdf5::node::Group &HDFGroup) {
...
+void HDFWriterModule::writeAdcPulseData(FlatbufferMessage const &Message) {
+  auto EventMsgFlatbuffer = GetEventMessage(Message.data());
+  if (EventMsgFlatbuffer->facility_specific_data_type() !=
+      FacilityData::AdcPulseDebug) {
+    return;
+  }
+  auto AdcPulseDebugMsgFlatbuffer =
+      EventMsgFlatbuffer->facility_specific_data_as_AdcPulseDebug();
+
+  auto AmplitudePtr = AdcPulseDebugMsgFlatbuffer->amplitude()->data();
```
SkyToGround: We are probably at a point where it is worth it to write a template function in order to minimise copy and paste code. Something similar to the following should probably work:
```
template <typename DataType>
ArrayAdapter<const DataType> getFBVectorAsArrayAdapter(const flatbuffers::Vector<DataType> *Data) {
  return {Data->data(), Data->size()};
}
```

src/tests/helpers/HDFFileTestHelper.h
```diff
@@ -15,7 +15,8 @@
 
 namespace HDFFileTestHelper {
 
-FileWriter::HDFFile createInMemoryTestFile(const std::string &Filename);
+FileWriter::HDFFile createInMemoryTestFile(const std::string &Filename,
```
SkyToGround: 👍

src/schemas/ev42/ev42_rw.cpp
```diff
@@ -200,31 +242,83 @@ HDFWriterModule::reopen(hdf5::node::Group &HDFGroup) {
...
+
+  if (RecordAdcPulseDebugData) {
+    writeAdcPulseData(Message);
+  }
+}
+void HDFWriterModule::writeAdcPulseData(FlatbufferMessage const &Message) {
+  auto EventMsgFlatbuffer = GetEventMessage(Message.data());
+  if (EventMsgFlatbuffer->facility_specific_data_type() !=
+      FacilityData::AdcPulseDebug) {
+    return;
```
SkyToGround: An unlikely (but not impossible) corner case is that some flatbuffers might not have the extra debug data whereas some does have it. A slightly safer implementation in that case would be to append the debug datasets with the appropriate number of zeroes if `RecordAdcPulseDebugData` is set to `true` but no elements are found in the relevant flatbuffer vectors.

However, due to how unlikely this is‚ I am fine with leaving the code as is.

src/tests/EventWriterTests.cpp
```diff
@@ -0,0 +1,440 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
SkyToGround: 👍

SkyToGround:
Btw, it seems likely to me that the build fails due to the file _conanfile_no_graylog.txt_ not having been updated to use the latest version of _streaming-data-types_.

src/schemas/ev42/ev42_rw.cpp
matthew-d-jones: Zero padding is now implemented and covered with unit test.

src/schemas/ev42/ev42_rw.cpp
matthew-d-jones: :+1: Done

src/schemas/ev42/ev42_rw.cpp
matthew-d-jones: For now I've converted from `chunk_bytes` to a number of elements so that it is clear what is going on in the code and the `chunk_bytes` option in the JSON will do what people expect.
When we refactor the rest of the module we could make the option `chunk_size` instead and avoid the problem.

src/schemas/ev42/ev42_rw.cpp
matthew-d-jones: Yes, I wasn't sure which to go for, in the end opted for the group because I would otherwise have to prefix the datasets with what I used as the group name. On the other hand being in the same group might be cleared that they should be consistent with `event_time_zero` and `event_index`.

matthew-d-jones:
> Btw, it seems likely to me that the build fails due to the file _conanfile_no_graylog.txt_ not having been updated to use the latest version of _streaming-data-types_.

I've now fixed that, but expect the build will still fail due to the weird compiler failure problems we are having on the build nodes.

src/schemas/ev42/ev42_rw.cpp
matthew-d-jones: As it was Tobias's suggestion I've removed the group. The ADC datasets now have a prefix of `adc_pulse_` (`adc_pulse_amplitude`, etc) and are in the `NXevent_data` group.

matthew-d-jones:
@SkyToGround Thanks for the review, I've addressed all comments.

src/schemas/ev42/ev42_rw.cpp
SkyToGround: Nice!

mkoennecke:
I wonder if this should at all be in the ev42 message and code. But rather a separate message and a separate writer. In order to separate concerns and declutter the code:
- here the ev42 main event data writing
- there the ADC debugging information writing

SkyToGround:
@matthew-d-jones I had not realised that the build failed. Hopefully, the compiler crashing was a temporary issue. I will have Jenkins do a re-build and if it is successful, I will merge the PR.

SkyToGround:
@mkoennecke Seeing as we need the functionality that Matt has implemented now, I will merge the code when possible. I suggest that you create a JIRA ticket with your comments.

PR #473, opened 2019-09-19
Improve command handling testing

mattclarke:
### Issue
N/A

### Description of work
Splits out some of the command handling into a separate class.
It's a step in the right direction, but there is still more to come.

### DIFF
--- a/src/CommandParser.h
+++ b/src/CommandParser.h
@@ -72,9 +72,9 @@ private:
   SharedLogger Logger = getLogger();
 
   uri::URI extractBroker(nlohmann::json const &JSONCommand);
-  std::string extractJobID(nlohmann::json const &JSONCommand);
+  static std::string extractJobID(nlohmann::json const &JSONCommand);
   static std::chrono::duration<long long int, std::milli> getCurrentTime();
-  std::chrono::milliseconds
+  static std::chrono::milliseconds
   extractTime(std::string const &Key, nlohmann::json const &JSONCommand,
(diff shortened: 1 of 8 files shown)


src/CommandParser.h
```diff
@@ -0,0 +1,104 @@
...
+
+  /// \brief Extract the command name from the command.
+  ///
+  /// Note: the command is converted to lower-case.
+  ///
+  /// \param JSONCommand The JSON Command.
+  /// \return The command name.
+  static std::string extractCommandName(const nlohmann::json &JSONCommand);
+
+  static std::string const StopCommand;
```
SkyToGround: Minor thing: Could these be private?

src/CommandParser.h
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
SkyToGround: It seems to me that having `CommandParser` as a class is not really justified. The only shared state is the pointer to the logger. I think that the implementation would be slightly cleaner (and easier to test) by making `CommandParser` into a namespace. The pointer to the logger could still be retained as a shared state by removing it from the header file and replacing it with `static SharedLogger Logger = getLogger();` in the *.cpp-file.

mattclarke: Agreed, that was something I was thinking off doing then forgot!

mattclarke: I'll see how the next set of changes go and see if they fit with this.

PR #474, opened 2019-09-19
Additions to NeXusDataset in preparation for future changes

SkyToGround:
### Issue

DM-1718


SkyToGround:
@matthew-d-jones I do see an argument for putting it in a separate file but I do not see an argument for putting it in the same directory as the ev42 module. The reason being that the dataset classes are supposed to represent NeXus datasets. Many of these are reusable across different (NX-)classes. I will refactor the code a bit now but feel free to give me suggestions on how you think the code should be divided up between files.

I admit that I was a bit lazy about writing unit tests for only these classes in isolation due to the minimal amount of code they added (an empty constructor).

matthew-d-jones:
I was referring to `EventTimeZero`, `EventIndex` and so on which you have declared in `NeXusDataset`, which is starting to get large. In this case they are specifically named as datasets which are only found in `NXevent_data`. I was thinking it would make sense to move them to `EventDataset.h` or something like that.
I guess having them in a central location makes sense. If we have to make a breaking change to `ev42` at some point, and thus have to create a new writer module, then we would want these dataset classes to be available to both modules.

SkyToGround:
I was thinking along the lines of your last sentence. I completely agree with _NeXusDataset.h_ getting to large and thus I split the code into _ExtensibleDataset.h_ and _NeXusDataset.h_. I also created some additional tests.

PR #475, opened 2019-09-23
Make command parser methods into free functions

mattclarke:
### Issue

N/A

### Description of work

As suggested by @SkyToGround I have made the methods into free functions.

I have also made it a requirement to include the broker details in the "start" command.
I don't think it makes much sense to have it default to localhost when not provided.

### DIFF
--- a/src/CommandParser.h
+++ b/src/CommandParser.h
@@ -16,6 +16,7 @@
 #include "URI.h"
 #include "json.h"
 #include "logger.h"
+#include "helper.h"
 #include <chrono>
 
 namespace FileWriter {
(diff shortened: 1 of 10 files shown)


src/CommandParser.h
```diff
@@ -38,67 +39,89 @@ struct StopCommandInfo {
...
-  static T getRequiredValue(std::string const &Key,
-                            nlohmann::json const &JSONCommand) {
-    if (auto x = find<T>(Key, JSONCommand)) {
-      return x.inner();
-    }
-
-    throw std::runtime_error(
-        fmt::format("Missing key {} from command JSON", Key));
+namespace CommandParser {
+static std::string const StartCommand = "filewriter_new";
```
SkyToGround: It is possible to hide these string constants in the *.cpp-file. Of course, this means that they are unavailable for (e.g.) unit testing. Thus, I am ok with leaving them as is.

matthew-d-jones: They are needed in `CommandHandler` I think.

SkyToGround: Ah, ok.

PR #478, opened 2019-09-25
Fix stop time functionality and allow writing of historic data

matthew-d-jones:
### Issue

Closes DM-1616 and DM-1148

### Description of work

Previous to these changes, the file writer stops consuming any data as soon as the current wallclock time passes the requested stop time. This means it was impossible to write old data, for example from an experiment run which has already finished. It also meant that if the file writer did not keep up with the data stream then we lost data from the end of the run. We observed both of these problems at V20.

The file writer now does the following:

- Until the system time >= stop time + max_producer_lag, continue to consume data but throw away any messages which are timestamped greater than stop time
- When system time >= stop time + max_producer_lag, get stop offsets using offsetsForTimes method.
- Continue to consume data until current offset >= stop offset, throwing away any messages which are timestamped greater than stop time

The new `test_filewriter_can_write_data_when_start_and_stop_time_are_in_the_past` system test demonstrates the new functionality. The functionality is also covered by unit tests in `StreamerTest.cpp`. There are several edge cases to be handled, for example that there is no data in the topic, that the stop offset was already reached when the stop offset is looked up, or that there is data in the topic but none between the start and stop offset. These are covered by unit tests.

During testing at V20 I discovered that I needed to improve the file writer's response to poor network conditions. To achieve this there are now retries on metadata lookups and I also updated librdkafka to 1.1.0 which is much more robust.


src/StreamMaster.cpp
```diff
@@ -150,6 +152,11 @@ void StreamMaster::closeStream(Streamer &Stream, const std::string &TopicName) {
         TopicName, NumStreamers);
   }
   Stream.close();
+
+  if (NumStreamers == 0) {
```
mattclarke: I removed this on a different PR.
I think for testing it is useful to be able to run up a stream master with no streams and not have it close itself.

src/Streamer.cpp
```diff
@@ -97,36 +97,112 @@ bool FileWriter::Streamer::ifConsumerIsReadyThenAssignIt() {
...
+/// offset for every partition
+/// \param NewMessagePartition partition number of the newly received message
+/// \param NewMessageOffset offset of the newly received message
+/// \return true if we've reached stop offset for every partition
+bool Streamer::stopOffsetsReached(int32_t NewMessagePartition,
+                                  int64_t NewMessageOffset) {
+  if (NewMessageOffset >= StopOffsets[NewMessagePartition].first) {
+    StopOffsets[NewMessagePartition].second = true;
+  }
+  return std::all_of(
```
mattclarke: Nice!

src/Streamer.cpp
```diff
@@ -97,36 +97,112 @@ bool FileWriter::Streamer::ifConsumerIsReadyThenAssignIt() {
...
+    CatchingUpToStopOffset = true;
+    Logger->trace("Calling getStopOffsets");
+    StopOffsets = getStopOffsets(Options.StartTimestamp,
+                                 Options.StopTimestamp + Options.AfterStopTime,
+                                 MessageProcessor.topic());
+    Logger->trace("Finished executing getStopOffsets");
+    // There may be no data in the topic, so check already if all stop offsets
+    // are marked as reached
+    bool NoDataInTopic =
+        std::all_of(StopOffsets.cbegin(), StopOffsets.cend(),
```
mattclarke: Duplicate code?

src/KafkaW/Consumer.cpp
```diff
@@ -63,115 +84,215 @@ Consumer::queryWatermarkOffsets(const std::string &Topic) {
...
+    findTopic(TopicName, KafkaMetadata);
   } catch (std::runtime_error &e) {
     return false;
-  };
+  }
   return true;
 }
 
 /// Updates the RdKafka Metadata pointer. If broker is unavailable, keeps trying
 /// to connect every 500ms, logging messages every few seconds.
```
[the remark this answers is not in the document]
matthew-d-jones: Thanks, updated to:
```
/// Gets metadata from the Kafka broker, if unsuccessful then keeps trying
/// and logs a warning message every WarnOnNRetries attempts.
```

src/KafkaW/Consumer.cpp
```diff
@@ -63,115 +84,215 @@ Consumer::queryWatermarkOffsets(const std::string &Topic) {
...
+  uint32_t LoopCounter = 0;
+  uint32_t WarnOnNRetries = 10;
+  std::shared_ptr<RdKafka::Metadata> KafkaMetadata = metadataCall();
+  while (KafkaMetadata == nullptr) {
+    if (LoopCounter == WarnOnNRetries) {
+      Logger->warn("Cannot contact broker, retrying until connection is "
+                   "established...");
       LoopCounter = 0;
     }
+    KafkaMetadata = metadataCall();
```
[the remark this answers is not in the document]
matthew-d-jones: I've never seen it fail any other way than timing out, and the code avoids spamming log messages by only warning every N attempts, so I think this is fine.

src/StreamMaster.cpp
matthew-d-jones: Currently if I remove this code then the writer would never stop and close the file... will discuss with you on Slack.

src/Streamer.cpp
matthew-d-jones: :+1: extracted method

PR #479, opened 2019-10-10
DM-1718, refactor file writing code. Part 2

SkyToGround:
### Issue

DM-1718

### Description of work

*Add a description of the work done. The aim is provide information to help the reviewer review the PR efficiently.
See [CONTRIBUTING.MD](CONTRIBUTING.MD) for more information on code standards etc.*


src/schemas/template/TemplateWriter.h
```diff
@@ -221,7 +218,7 @@ class WriterClass : public FileWriter::HDFWriterModule {
   /// \param HDFAttributes Additional attributes as defined in the Nexus
   /// structure which the HDFWriterModule should write to the file. Because the
   /// HDFWriterModule is free to create the structure and datasets according to
-  /// its needs, it must also take the responsibility to write these
+  /// its needs, it must also take the reposnsibility to write these
```
mattclarke: Why did you add a typo?

mattclarke: ;)

SkyToGround: Good question. I do not know.

SkyToGround:
```suggestion
  /// its needs, it must also take the responsibility to write these
```

src/HDFWriterModule.h
```diff
@@ -9,16 +9,16 @@
 
 #pragma once
 
+#include "FlatbufferMessage.h"
```
matthew-d-jones: You don't like forward declarations? They can drastically decrease compile times...

src/Source.cpp
```diff
@@ -8,6 +8,11 @@
 // Screaming Udder!                              https://esss.se
 
 #include "Source.h"
+#include "helper.h"
+#include "logger.h"
+#include <chrono>
+#include <fstream>
+#include <thread>
```
matthew-d-jones: Why have all these includes appeared? It doesn't look like any code was added which need them.

src/Source.h
```diff
@@ -12,7 +12,9 @@
 #include "FlatbufferReader.h"
 #include "HDFFile.h"
 #include "HDFWriterModule.h"
+#include "Msg.h"
 #include "ProcessMessageResult.h"
+#include "json.h"
```
matthew-d-jones: Why the extra includes?

src/HDFWriterModule.h
SkyToGround: It's more about not actively searching for places where they can be used. I can add one in this case but if you know some tool that automates finding opportunities for forward declarations, that would be great.

matthew-d-jones: I asked because it looks like you actively removed one here.
When writing new stuff I think it is just a case of being aware of the cost of adding includes to headers: forward declare by default and only `#include` if that doesn't work.

SkyToGround: I can't be bother to figure out exactly why it happened but seeing as you added this forward declarations to this file very recently, it seems likely that the reason for it disappearing was because I branched off of master before that change had been merged.

SkyToGround: Considering the changes I had to do to make the forward declaration work, my previous comment seems like a very likely reason.

src/Source.cpp
SkyToGround: I do not know but it is not entirely unlikely that the reason is something similar to the comment above this one. It has now been fixed.

src/Source.h
SkyToGround: Ditto.

SkyToGround:
@matthew-d-jones Thanks for the review! I have fixed the issues that you pointed out.

SkyToGround:
Looks like I have resolved the merge conflicts.

PR #480, opened 2019-10-14
Detangle command handler

mattclarke:
### Issue

N/A

### Description of work

Breaking stuff out of command handler, so it at least has some semblance of SRP about it.
Mostly a case of  moving code around.

There are a number of other things that could be done but this is probably a good place to stop with [DM-1730](https://jira.esss.lu.se/browse/DM-1730) on the horizon.

Need a better name for CommandHandler - any suggestions?


src/tests/helpers/FakeStreamMaster.h
```diff
@@ -0,0 +1,29 @@
...
+  void requestStop() override { IsRemovable = true; }
+  bool isRemovable() const override { return IsRemovable; }
+  void setStopTime(const std::chrono::milliseconds & /*StopTime*/) override {}
+
+  nlohmann::json getStats() const override { return nlohmann::json::object(); }
+
+private:
+  std::string JobID;
+  bool IsRemovable;
+};
```
SkyToGround:
```suggestion
};

```

src/JobCreator.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include <chrono>
+
+using std::vector;
+
+namespace FileWriter {
+
+using nlohmann::json;
+
+std::vector<StreamHDFInfo>
+JobCreator::initializeHDF(FileWriterTask &Task,
```
SkyToGround: 👍

src/JobCreator.cpp
```diff
@@ -0,0 +1,240 @@
...
+                    StreamHDFInfo.HDFParentName, StreamHDFInfo.ConfigStream);
+    }
+  }
+  return StreamSettingsList;
+}
+
+std::unique_ptr<IStreamMaster> JobCreator::createFileWritingJob(
+    StartCommandInfo const &StartInfo,
+    std::shared_ptr<KafkaW::ProducerTopic> const &StatusProducer,
+    MainOpt &Settings) {
```
SkyToGround: I do not request a change as such here but I do have a question: Do we break the code if we turn `MainOpt &Settings` into `MainOpt Settings` (i.e. remove the reference)?

src/JobCreator.cpp
```diff
@@ -0,0 +1,240 @@
...
+               Settings.StreamerConfiguration.StartTimestamp.count());
+  if (Settings.StreamerConfiguration.StopTimestamp.count() > 0) {
+    Logger->info("Stop time: {}ms",
+                 Settings.StreamerConfiguration.StopTimestamp.count());
+  }
+
+  Logger->info("Write file with job_id: {}", Task->jobID());
+  auto s = StreamMaster::createStreamMaster(
+      StartInfo.BrokerInfo.HostPort, std::move(Task), Settings, StatusProducer);
+  s->report(std::chrono::milliseconds{Settings.StatusMasterIntervalMS});
```
SkyToGround: General comment: We should probably convert `StatusMasterIntervalMS` into a `std::chrono::duration`.

src/Master.cpp
```diff
@@ -59,27 +53,65 @@ void Master::handle_command_message(std::unique_ptr<Msg> CommandMessage) {
...
+  }
 }
 
-std::unique_ptr<StreamMaster> &
-Master::getStreamMasterForJobID(std::string const &JobID) {
-  for (auto &StreamMaster : StreamMasters) {
-    if (StreamMaster->getJobId() == JobID) {
-      return StreamMaster;
+void Master::handle_command(std::string const &Command,
+                            std::chrono::milliseconds TimeStamp) {
```
SkyToGround: I do not see that there is a good reason to move the timestamp around. It looks like it is only used in one location to me. Maybe get the current timestamp there?

src/JobCreator.cpp
mattclarke: No, I don't think so. It certainly builds at least

src/JobCreator.cpp
mattclarke: Done!

PR #483, opened 2019-11-20
Change NICOS cache writer to conform to NXLog rules

mattclarke:
### Issue

DM-1375

### Description of work

NICOS cache values are now written to NeXus as doubles. If the value cannot be cast then it is ignored. In the future we should look to move to the typed ns11 schema.
Added some tests and tidied some minor things.

### DIFF
--- a/src/schemas/ns10/NicosCacheWriter.h
+++ b/src/schemas/ns10/NicosCacheWriter.h
@@ -40,7 +40,7 @@ public:
 protected:
   std::string Sourcename;
   NeXusDataset::DoubleValue Values;
-  hdf5::Dimensions ChunkSize{64};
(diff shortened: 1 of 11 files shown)


src/schemas/ns10/NicosCacheWriter.h
```diff
@@ -48,7 +39,7 @@ class CacheWriter : public FileWriter::HDFWriterModule {
 
 protected:
   std::string Sourcename;
-  StringValue Values;
+  NeXusDataset::DoubleValue Values;
   hdf5::Dimensions ChunkSize{64};
```
SkyToGround:
```suggestion
  hdf5::Dimensions ChunkSize{1024};
```

PR #488, opened 2019-12-12
Inject demux into streamer

mattclarke:
### Issue

N/A

### Description of work

Added the DemuxTopic via the constructor.
And re-added the tests for Streamer.


mattclarke:
I am just seeing if I can work out why that one system test is failing. If I don't get it working soon then I think we can just merge it

PR #489, opened 2019-12-17
Broke the large processing methods up.

mattclarke:
### Issue

Work towards DM-1730

### Description of work

Broke up the large methods and, hopefully, produced an obvious 'flow'.

### DIFF
--- a/src/Streamer.cpp
+++ b/src/Streamer.cpp
@@ -222,7 +222,7 @@ bool Streamer::messageHasPayload(KafkaW::PollStatus MessageStatus) {
 
 bool Streamer::messageSourceIsValid(size_t SourceHash) const {
   return !(MessageProcessor->sources().find(SourceHash) ==
-           MessageProcessor->sources().end());
+      MessageProcessor->sources().end());
 }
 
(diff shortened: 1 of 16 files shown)


src/Streamer.cpp
```diff
@@ -199,76 +197,88 @@ void Streamer::processMessage(
...
   }
 
-  if (KafkaMessage->first == KafkaW::PollStatus::Empty ||
-      KafkaMessage->first == KafkaW::PollStatus::EndOfPartition ||
-      KafkaMessage->first == KafkaW::PollStatus::TimedOut) {
-    return;
+  return false;
+}
+
+bool Streamer::kafkaMessageIsValid(KafkaW::PollStatus MessageStatus) {
```
matthew-d-jones: I'm not too keen on the name of this function because it sounds to me like we are checking the contents of the message payload. Really we are checking whether an actual message payload has been returned at all. Maybe `messageHasPayload()`?

mattclarke: 👍 Makes sense - thanks

PR #491, opened 2019-12-18
Simplify system tests command files

mattclarke:
### Issue

N/A

### Description of work

Removed some of the command files by using a generic stop command file and getting the job id from the start command.

A number of the start command files could also be removed by supplying start and stop times, and file names from the method. 

May make sense to break the send_writer_command into two separate start and stop commands for clarity. Not sure.


system-tests/test_filewriter_stop_time.py
```diff
@@ -33,20 +35,20 @@ def test_filewriter_clears_stop_time_between_jobs(docker_compose_stop_command):
...
         if (
             '"code":"START"' in message
-            and '"job_id":"a8e31c99-8df9-4123-8060-2e009d84a0df"' in message
+            and f'"job_id":"{job_id}"' in message
         ):
             started = True
         if (
             '"code":"CLOSE"' in message
-            and '"job_id":"a8e31c99-8df9-4123-8060-2e009d84a0df"' in message
+            and f'"job_id":"{job_id}"' in message
```
rerpha: should possibly state in the systests readme that python 3.6 or above is required if format strings are to be used

rerpha: personally i think we should make this a requirement anyway, and developing or running anything under python 3.6 should be appended to the 7 deadly sins

PR #499, opened 2020-01-09
Dm 1799 new metrics system mdj tweaks

matthew-d-jones:
Functionality should be identical to Jonas's implementation in `DM-1799_add_new_metrics`, this is just refactored into a slightly different design.
`Processor` is replaced by `Reporters`, one per different metrics `Sink` (logger, carbon etc). Interfacing with a different database, for example InfluxDB, would simply require implementing a new `Sink` and tweaking the `Reporter` factory function. Each `Reporter` is independent, so different `Sinks` could be configurable via the optargs.

Not ready for merging. The draft PR is up to invite feedback on the design.


src/Metrics/Registrar.h
```diff
@@ -1,32 +1,37 @@
...
+/// Register and deregister metrics to be reported via a specified sink
+/// threadsafe
 class Registrar {
 public:
-  Registrar(Registrar &&Registrar);
-
-  bool registerMetric(Metric &NewMetric, DestList Destinations);
-
-  Registrar getNewRegistrar(std::string const &MetricsPrefix) const;
+  Registrar() = default;
```
SkyToGround: This design of `Registrar` now requires the class to know of both "sinks" and metrics list when previously it did not. I do not think it is an improvement.

src/Metrics/Metric.h
```diff
@@ -2,60 +2,52 @@
...
-using CounterType = std::atomic<std::int64_t>;
-using InternalCounterType = decltype(((CounterType *)(nullptr))->load());
+using CounterType = std::atomic<int64_t>;
 
 class Metric {
 public:
-  Metric(std::string Name, std::string Description,
-         Severity Level = Severity::DEBUG)
-      : MName(std::move(Name)), MDesc(std::move(Description)), SevLvl(Level) {}
+  Metric(std::shared_ptr<Registrar> RegistrarUsed, std::string Name,
```
SkyToGround: Providing a registrar to a Metrics::Metric constructor is bad interface design if we are also requiring the user to call Registrar::registerMetric. Have one alternative, not both.
Furthermore, the current implementation requires us to (indirectly or directly) hold on to the registrar after we are done with it.

src/Metrics/ReporterFactory.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Metrics {
+
+class Reporter;
+class Registrar;
+enum struct LogTo;
+
+/// Create a Reporter with a specified type of Sink
+std::unique_ptr<Reporter>
+createReporter(std::shared_ptr<Registrar> const &MetricsRegistrar,
```
SkyToGround: I do not see a good reason to have this factory.

matthew-d-jones: It is intended to tidy away creation of the `MetricsList` which is then shared between the `Registrar` and the `Reporter`. It is not really factory pattern, so should at least be renamed.

src/Metrics/Registrar.h
matthew-d-jones: It previously had to know about the `Processor` instead of metrics list. It doesn't need to know about "sinks", just the enum for sinks, which it also had before. I think I can resolve that.

src/Metrics/Metric.h
matthew-d-jones: Agree on the first point, should be easily resolved.
The second point is true, and longer object lifetime is generally bad. I'll have a think about it.

matthew-d-jones:
@SkyToGround Thanks for the comments, I had spotted there are some missing lock_guards and related issues when updating the tests.
Not sure about `Sink` naming, it is pretty standard terminology for that kind of thing and it is in a different namespace. It doesn't result in any duplicate class or file names and confusing `spdlog::sinks` with `Metrics::Sink` doesn't seem so likely to me. But got another name suggestion?

SkyToGround:
My original idea of what the interface should (roughly) look like:

```
struct B {
  B(Registrar &Reg) {
    auto UsedRegistrar = Reg.getRegistrar("class_B");
    UsedRegistrar.registerMetric(SomeOtherMetric); // some_app.class_A.calss_B.some_metric_name
  }
  Metric SomeOtherMetric{"some_metric_name", "desc"};
};

struct A {
  A(auto& Registrar) : InstanceB(Registrar.getRegistrar("class_A")) {
    auto TempReg = Registrar.getRegistrar("class_A");
    TempReg.registerMetric(SomeMetric); // some_app.class_A.some_metric_name
  }
  B InstanceB;
  Metric SomeMetric{"some_metric_name", "desc"};
};


int main() {
  Processor SomeProcessor("some_app", "localhost", 1234, 1s, 1s);
  
  A InstanceA(SomeProcessor.getRegistrar(););
}
```

matthew-d-jones:
To summarise what we discussed over VC:
Reverting to the thing doing the reporting (formerly `Processor` now `Reporter`) having sole ownership of the list of metrics to report on, and the `Registrar` simply being the name prefix-management layer, should resolve the issues @SkyToGround has pointed out.

SkyToGround:
@matthew-d-jones Yes. I would also like to point out that seeing as the reporter uses ASIO for timing purposes anyway, you could remove all (visible) mutexes by doing registration (and de-registration) in a way that is similar to this:
```
void Reporter::registerMetric(Metric& NewMetric) {
  NewMetric.registerReporter(this);
  IO.post([=](){
    MetricsList.emplace_back(NewMetric);
  });
```

src/Metrics/Reporter.h
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  void addMetric(Metric &NewMetric, std::string const &NewName) {
+    IO.post([&]() {
+      MetricsToReportOn.emplace(NewName, InternalMetric(NewMetric, NewName));
+    });
+  }
+
+  void tryRemoveMetric(std::string const &MetricName) {
+    IO.post([&]() { MetricsToReportOn.erase(MetricName); });
```
SkyToGround: For removing a metric, this has to be a blocking call as the metrics thread might access the counter after it has been deallocated by the `Metric` instance. I have successfully used code similar to the following:
```
void tryRemoveMetric(...) {
  auto WorkDone = std::make_shared<std::promise<void>>();
  auto WorkDoneFuture = WorkDone->get_future();
  IO.post([&, WorkDone{std::move(WorkDone)} ]() {
    MetricsToReportOn.erase(MetricName);
    WorkDone->set_value();
  });
  WorkDoneFuture.wait();
}
```

src/Metrics/Reporter.h
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  LogTo getSinkType() { return MetricSink->getType(); };
+
+  void start() {
+    AsioTimer.async_wait(
```
SkyToGround: Any reason not to do this in the constructor?

src/Metrics/Reporter.h
```diff
@@ -0,0 +1,68 @@
...
+#include <thread>
+
+namespace Metrics {
+
+class Reporter {
+public:
+  Reporter(std::unique_ptr<Sink> MetricSink, std::chrono::milliseconds Interval)
+      : MetricSink(std::move(MetricSink)), IO(), Period(Interval),
+        AsioTimer(IO, Period){};
+
```
SkyToGround: We should have a destructor that automatically stops the thread. It is too easy to forget to call `waitForStop()`.

src/Metrics/Reporter.h
```diff
@@ -0,0 +1,72 @@
...
+  virtual ~Reporter() = default;
```
SkyToGround: We should print an error if there is one or more metric that has not been de-registered when calling the destructor.

matthew-d-jones: Do we care if `Reporter` doesn't outlive all `Metric`s

SkyToGround: If this happens, then it is an indication that there likely is a serious bug in the code.

matthew-d-jones:
@SkyToGround There is likely one CppCheck error remaining because `Registrar::getNewRegistrar()` isn't yet used outside of tests, but presumably the same error exists on your branch (which this will merge into) anyway.

src/Metrics/CarbonSink.cpp
```diff
@@ -0,0 +1,15 @@
+#include "CarbonSink.h"
+#include "InternalMetric.h"
+
+namespace Metrics {
+
+void CarbonSink::reportMetric(InternalMetric &MetricToBeReported) {
+  auto CurrentValue = MetricToBeReported.Counter->load();
```
SkyToGround: We are mixing memory ordering here (we are using relaxed memory ordering elsewhere). I am assuming that this has not effect on atomicity (is that a word?), do you know?

src/Metrics/Reporter.cpp
```diff
@@ -0,0 +1,66 @@
...
+      [this](std::error_code const & /*error*/) { this->reportMetrics(); });
+  ReporterThread = std::thread(&Reporter::run, this);
+}
+
+void Reporter::waitForStop() {
+  // AsioTimer.cancel() would only stop the timer execution if there is an
+  // async_wait "in flight" we therefore need the Running flag and supporting
+  // logic too, to ensure that the reportMetrics call chain is definitely
+  // stopped
+  Running = false;
```
SkyToGround: Instead of having the `Running` variable, you should simply call `IO.stop()`.

src/Metrics/Reporter.cpp
```diff
@@ -0,0 +1,66 @@
...
+  AsioTimer.cancel();
+  ReporterThread.join();
+}
+
+Reporter::~Reporter() {
+  if (!MetricsToReportOn.empty()) {
```
SkyToGround: We are missing a lock guard here.
```suggestion
  std::lock_guard<std::mutex> Lock(MetricsMapMutex);
  if (!MetricsToReportOn.empty()) {
```

src/Metrics/Reporter.cpp
matthew-d-jones: :+1: Ah of course, we can do that in this case because IO is only being used for this timer.

src/Metrics/CarbonSink.cpp
matthew-d-jones: My guess is that this would force sequential everywhere. No reason not to continue with relaxed, so I'll specify it here (and in the log sink) and add a note to the Sink interface for people adding new sinks in the future.

PR #503, opened 2020-01-20
Create breaking_changes.md

matthew-d-jones:
### Issue

Dm-1819

### Description of work

Added a markdown file to keep track of breaking changes to document in release notes.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -11,3 +11,6 @@ See [CONTRIBUTING.MD](CONTRIBUTING.MD) for more information on code standards et
 
 - [ ] Nominate for code review 
 
+### Reminder
+
+*Breaking changes should be documented in `breaking-changes.md`*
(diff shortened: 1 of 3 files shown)


matthew-d-jones:
@SkyToGround 
1. I don't see any way other than trying to make it habit to add to the changes file and for the PR reviewer to check that has been done. This is the approach the Mantid project (successfully) takes.
3. Good idea. If we document all changes this way then that also solves any issue with identifying a "breaking change".

Overall this is an experimental change to methodology, in the spirit of agile development. We should review whether it has been effective or should be altered/replaced/scrapped once we've done the next release.

mattclarke:
Yeah, I agree we should track all changes.
It might require us to change how we organise our branches etc with regards to version numbers.

I think it is fair to say we are working on version 2.0.0, so it works in this instance

matthew-d-jones:
It does practically guarantee a merge conflict on every branch unfortunately, but at least it is an easily resolved one.

PR #506, opened 2020-01-27
Dm 996 part3

SkyToGround:
### Description of work

* Added the functionality for multiple modules for one FB-id.
* Added/fixed tests.
* Created system test (is disabled by default).


Dockerfile
```diff
@@ -30,10 +30,10 @@ COPY ./CMakeLists.txt ../kafka_to_nexus_src/CMakeLists.txt
 
 RUN cd kafka_to_nexus \
     && cmake -DCONAN="MANUAL" --target="kafka-to-nexus" -DCMAKE_BUILD_TYPE=Release -DUSE_GRAYLOG_LOGGER=True -DRUN_DOXYGEN=False -DBUILD_TESTS=False ../kafka_to_nexus_src \
-    && make -j8 kafka-to-nexus \
+    && make kafka-to-nexus \
```
matthew-d-jones: Jack found that using `-j8` was not noticeably detrimental to build time on the system test jenkins nodes (which have 2 threads, so we should at least use `-j2`) and significantly improved build time on dev machines.
Given that `ninja` is small and clever enough to figure out how to parallelise itself appropriately, an alternative would be to use that instead of `make`.

system-tests/test_same_id_multiple_modules.py
```diff
@@ -0,0 +1,28 @@
+from helpers.kafkahelpers import create_producer, send_writer_command
+from helpers.nexushelpers import OpenNexusFileWhenAvailable
+from time import sleep
+import numpy as np
+import pytest
+
+
+@pytest.mark.skip(reason="Is painfully slow to run")
+def test_data_reaches_file(docker_compose):
```
matthew-d-jones: These are not unit tests. I suggest modifying the existing test in `test_filewriter.py rather than creating a new test.

Dockerfile
SkyToGround: Fair enough. When running it on my machine I found that single threaded compilation was noticeable faster than `-j8`. If course, I did not test `-j2`.

SkyToGround:
```suggestion
    && make -j2 kafka-to-nexus \
```

system-tests/test_same_id_multiple_modules.py
SkyToGround: I can do that but I would like you to give me a rule of thumb for when to put the system tests in a separate file given that the files _test_filewriter_lr.py_, _test_filewriter.py_, _test_filewriter_ignores_commands.py_ and _test_filewriter_static_data.py_ only have one test each.

matthew-d-jones: I'm not suggesting moving your test to the existing file, I'm suggesting you modify the command for the existing test so that it also covers what your new test covers.
Broadly different scenarios that would require starting the file writer with different configurations warrant a new test, otherwise modifying an existing test is usually best. We can't have a separate system test for every unit of functionality, that is what unit tests are for.

SkyToGround: Ah, I understand what you mean now. This test tests a slightly different code path than the one in `test_filewriter.py`. I could in principle modify that test to also execute this code path but I am not going to because how painful it currently is to:

1. Write/modify/test system tests.
2. Modify/add the data that is pushed to the kafka broker.

Given your last comment, I see four reasonable alternatives here:

1. Do nothing.
2. Delete _test_same_id_multiple_modules.py_.
3. Delete _test_filewriter.py_.
4. Have someone other than me (e.g. you) combine the two tests.

system-tests/test_same_id_multiple_modules.py
```diff
@@ -0,0 +1,28 @@
+from helpers.kafkahelpers import create_producer, send_writer_command
+from helpers.nexushelpers import OpenNexusFileWhenAvailable
+from time import sleep
+import numpy as np
+import pytest
+
+
+@pytest.mark.skip(reason="Is painfully slow to run")
```
SkyToGround:
```suggestion
```

system-tests/test_same_id_multiple_modules.py
SkyToGround: By the way, I am barely ok with option number 2 as I now know (because of the system test) that the code works. However, the code path executed in this case is somewhat important and should really be tested in case of future regression.

PR #516, opened 2020-02-21
Update run start stop messages

matthew-d-jones:
### Issue

DM-1834

### Description of work

Start/stop messages are now flatbuffers rather than JSON.
Updated unit tests and system tests.

I initially added verification of the flatbuffers, but there seem to be problems with verifying flatbuffers constructed with the python library. For now I've disabled the relevant unit tests and added a comment in the code to explain why we are not verifying.


src/FlatbufferMessage.h
```diff
@@ -54,7 +54,7 @@ class FlatbufferMessage {
   ///
   /// \param BufferPtr Pointer to memory location containing flatbuffer.
   /// \param Size Size of flatbuffer in bytes.
-  FlatbufferMessage(char const *BufferPtr, size_t Size);
+  FlatbufferMessage(uint8_t const *BufferPtr, size_t Size);
```
SkyToGround: Can you please motivate this change? I.e. what do we gain and why is it better than what we had before.

matthew-d-jones: I added ~4 reinterpret casts and removed ~20. `uint8_t` is what flatbuffers uses internally.

SkyToGround: 👍

src/tests/CommandParserTests.cpp
```diff
@@ -7,318 +7,261 @@
...
   auto FakeCurrentTime = std::chrono::milliseconds{987654321};
 
   auto StartInfo = FileWriter::CommandParser::extractStartInformation(
-      nlohmann::json::parse(Command), FakeCurrentTime);
+      MessageBuffer, FakeCurrentTime);
 
   ASSERT_EQ(FakeCurrentTime, StartInfo.StartTime);
 }
 
-TEST(CommandParserStartTests, IfNoStopTimeThenSetToZero) {
```
SkyToGround: There is too much code for me to review in detail but I would like to have you comment on if you have replaced every unit tests that you removed with a new one or if one or more unit tests were removed entirely.

src/tests/MasterTests.cpp
```diff
@@ -7,40 +7,30 @@
...
-    "file_attributes": {"file_name": "output_file1.nxs"},
-    "nexus_structure": { }
-  })"""};
-
-std::string StopCommand{R"""({
-    "cmd": "filewriter_stop",
-    "job_id": "1234"
-  })"""};
-
-TEST(ParseCommandTests, IfCommandStringIsParseableThenDoesNotThrow) {
```
SkyToGround: Ditto about replacement of unit tests.

src/tests/CommandParserTests.cpp
matthew-d-jones: Flatbuffers has no concept of "NoStopTime", the field is always present and therefore it will always give a value. So this test no longer made sense.

src/tests/MasterTests.cpp
matthew-d-jones: The nearest equivalent now would be passing verification, which we can't do for reasons explained here https://github.com/ess-dmsc/kafka-to-nexus/blob/ef529d1bbe46c8f6aa3d216e562d7d62f827c114/src/CommandParser.cpp#L102-L104. So this test was removed.

system-tests/config-files/file_writer_config_static_data.ini
```diff
@@ -2,6 +2,5 @@ hdf-output-prefix="/output-files/"
 command-uri="localhost:9092/TEST_writerCommand"
 status-uri="localhost:9092/TEST_writerStatus"
 status-master-interval=2000
-use-hdf-swmr=false
```
SkyToGround: Though a small change, it might significant. Can you comment on it?

src/tests/MasterTests.cpp
matthew-d-jones: The two tests you have commented on, plus the "successful parsing" version of this test are the only ones removed. All the others were just updated to take flatbuffer input rather than a JSON string.

src/tests/CommandParserTests.cpp
SkyToGround: 👍

src/tests/MasterTests.cpp
SkyToGround: Thanks.

system-tests/config-files/file_writer_config_static_data.ini
matthew-d-jones: I removed the system test which used the NeXus-Streamer because it is no longer covering anything that the other tests don't cover. This change helps maintains a balance between tests which use SWMR and which don't.

SkyToGround: Ok.

src/fb_metadata_extractors/f142/f142_Extractor.cpp
```diff
@@ -23,8 +21,8 @@ bool f142_Extractor::verify(FlatbufferMessage const &Message) const {
 /// Extract name of source from the message
 std::string
 f142_Extractor::source_name(FlatbufferMessage const &Message) const {
-  auto fbuf = get_fbuf(Message.data());
-  auto s1 = fbuf->source_name();
+  auto LogDataBuffer = GetLogData(Message.data());
```
mattclarke: These sort of things can all be `const`?

src/fb_metadata_extractors/f142/f142_Extractor.cpp
```diff
@@ -23,8 +21,8 @@ bool f142_Extractor::verify(FlatbufferMessage const &Message) const {
...
+  auto s1 = LogDataBuffer->source_name();
```
mattclarke: Can you fix the name while you are here? ;)

PR #524, opened 2020-03-13
Implemented message writer thread + other required minor changes

SkyToGround:
### Issue

* Several, can't be bothered to list them right now.

### Notes

* I had to make some minor changes to a few existing classes to make this work.
* Despite me saying that we should not be using ASIO, I could not help but implement the writer thread using a light version of ASIO that is called `ThreadedExecutor`. It reduces the amount of code that I need in the actual writer implementation. It also makes it easier (for some definitions of easier) to unit test. However, it could also be argued that it makes the code harder to reason about. Please leave your thoughts about this. It is a relatively simple matter to replace it with a regular loop if you think so.


src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+#include <functional>
+#include <future>
+#include <memory>
+#include <thread>
+
+class ThreadedExecutor {
+private:
+public:
+  using WorkMessage = std::function<void()>;
+  ThreadedExecutor(bool LowPrioThreadExit = false)
```
matthew-d-jones:
```suggestion
  explicit ThreadedExecutor(bool LowPrioThreadExit = false)
```

matthew-d-jones: Unless they are in common use, I think using abbreviations only hurts readability. Would prefer `Priority` over `Prio`.

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+private:
+  bool RunThread{true};
+  std::function<void()> ThreadFunction{[=]() {
+    while (RunThread) {
+      WorkMessage CurrentMessage;
+      if (MessageQueue.try_dequeue(CurrentMessage)) {
+        CurrentMessage();
+      } else if (LowPrioMessageQueue.try_dequeue(CurrentMessage)) {
```
matthew-d-jones: What's the idea for the priority levels here, I don't think we discussed it?
It isn't used in tests etc makes me particularly nervous.

SkyToGround: To be fair, it is a bit of a bodge. The reason is that `moodycamel::ConcurrentQueue` does not give **ANY** guarantees about ordering of items produced from multiple threads. Thus, to prevent the thread from exiting before everything is written to file, a low priority message queue is used.

This is one reason to forgo the use of `ThreadedExecutor`.

SkyToGround: By the way: Seeing as I know that you rarely read the PR-description: please comment on the use of `ThreadedExecutor`.

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+#include <concurrentqueue/concurrentqueue.h>
+#include <functional>
+#include <future>
+#include <memory>
+#include <thread>
+
+class ThreadedExecutor {
+private:
+public:
+  using WorkMessage = std::function<void()>;
```
matthew-d-jones: Maybe `WorkMessage` -> `Job`, it is less overloaded and standard terminology?

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+      : LowPrioExit(LowPrioThreadExit), WorkerThread(ThreadFunction) {}
+  ~ThreadedExecutor() {
+    if (LowPrioExit) {
+      SendLowPrioWork([=]() { RunThread = false; });
+    } else {
+      SendWork([=]() { RunThread = false; });
+    }
+    WorkerThread.join();
+  }
+  void SendWork(WorkMessage Message) { MessageQueue.enqueue(Message); }
```
matthew-d-jones:
```suggestion
  void SendWork(WorkMessage const &Message) { MessageQueue.enqueue(Message); }
```

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+  size_t size_approx() { return MessageQueue.size_approx(); }
```
matthew-d-jones: Method is not used.
```suggestion

```

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+  void SendLowPrioWork(WorkMessage Message) {
```
matthew-d-jones:
```suggestion
  void SendLowPrioWork(WorkMessage const &Message) {
```

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
+/* Copyright (C) 2018 European Spallation Source, ERIC. See LICENSE file */
```
matthew-d-jones: Doesn't have the full header comment?

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,58 @@
...
+      } else if (LowPrioMessageQueue.try_dequeue(CurrentMessage)) {
+        CurrentMessage();
+      } else {
+        using namespace std::chrono_literals;
+        std::this_thread::sleep_for(5ms);
+      }
+    }
+  }};
+  moodycamel::ConcurrentQueue<WorkMessage> MessageQueue;
+  moodycamel::ConcurrentQueue<WorkMessage> LowPrioMessageQueue;
```
matthew-d-jones: Wasn't obvious to me so I think worth explaining the role of the low priority queue in a comment here or in class description comment.

src/ThreadedExecutor.h
matthew-d-jones: I'm happy to proceed with this design. A little concerned with possible overhead of `std::function` for every message, but I suggest we do some performance review, profiling etc of the v3 branch before it hits master.

src/Stream/Message.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include "Message.h"
+
+namespace Stream {
+
+Message::Message(WriterModule::Base *DestinationModule,
+                 FileWriter::FlatbufferMessage const &Msg)
+    : FbMsg(Msg), DestPtr(DestinationModule) {}
+
+} // namespace Stream
```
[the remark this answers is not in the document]
SkyToGround:
```suggestion
} // namespace Stream

```

src/Stream/Message.h
```diff
@@ -0,0 +1,33 @@
...
+#include "FlatbufferMessage.h"
+#include <memory>
+
+namespace WriterModule {
+class Base;
+}
+
+namespace Stream {
+
+class Message {
```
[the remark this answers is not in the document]
SkyToGround: My thinking is that you should read it as `Stream::Message` which helps a bit. `MessageForModule` does not help here I think as it is not obvious what a module is. @mattclarke , do you want to help out here?

src/Stream/MessageWriter.cpp
```diff
@@ -0,0 +1,72 @@
...
+  Registrar.registerMetric(*ModuleErrorCounters[UnknownModuleHash],
+                           {Metrics::LogTo::LOG_MSG});
+}
+
+void MessageWriter::addMessage(Message Msg) {
+  Executor.SendWork([=]() { writeMsgImpl(Msg.DestPtr, Msg.FbMsg); });
+}
+
+void MessageWriter::writeMsgImpl(WriterModule::Base *ModulePtr,
+                                 FileWriter::FlatbufferMessage const Msg) {
```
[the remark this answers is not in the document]
SkyToGround: Right, done.

src/Stream/MessageWriter.cpp
```diff
@@ -0,0 +1,72 @@
...
+  Registrar.registerMetric(WritesDone, {Metrics::LogTo::CARBON});
+  Registrar.registerMetric(WriteErrors,
+                           {Metrics::LogTo::CARBON, Metrics::LogTo::LOG_MSG});
+  ModuleErrorCounters[UnknownModuleHash] = std::make_unique<Metrics::Metric>(
+      "error_unknown", "Unknown flatbuffer message.", Metrics::Severity::ERROR);
+  Registrar.registerMetric(*ModuleErrorCounters[UnknownModuleHash],
+                           {Metrics::LogTo::LOG_MSG});
+}
+
+void MessageWriter::addMessage(Message Msg) {
```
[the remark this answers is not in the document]
SkyToGround: Yes.

src/ThreadedExecutor.h
SkyToGround: Done.

src/ThreadedExecutor.h
SkyToGround: Ok, thanks!

src/Stream/Message.cpp
```diff
@@ -0,0 +1,18 @@
...
+//
+// Screaming Udder!                              https://esss.se
+
+#include "Message.h"
+
+namespace Stream {
+
+Message::Message(WriterModule::Base *DestinationModule,
+                 FileWriter::FlatbufferMessage const &Msg)
+    : FbMsg(Msg), DestPtr(DestinationModule) {}
```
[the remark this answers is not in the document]
SkyToGround: 1. Ok.
2. Yes, will do so.

src/ThreadedExecutor.h
SkyToGround: Does not sound (look) right to me when trying it. How about `JobType` and then rewrite `SendWork` into `SendWork(JobType Task)`?

src/ThreadedExecutor.h
SkyToGround: Ok.

src/ThreadedExecutor.h
SkyToGround: In this case, I think the use of `std::move` is the better solution.

src/ThreadedExecutor.h
SkyToGround: Deleted.

src/ThreadedExecutor.h
SkyToGround: Copied the file from another project that uses another style of header comment. Replaced it with one used here.

src/ThreadedExecutor.h
SkyToGround: Will do.

src/ThreadedExecutor.h
SkyToGround: Made the same change as I did for `SendWork`.

src/FlatbufferMessage.h
```diff
@@ -9,56 +9,90 @@
...
   /// Constructor is used in unit testing code to simplify set-up.
   FlatbufferMessage() = default;
 
-  /// \brief Verifies the data in the flatbuffer to make sure if it is valid.
+  /// \brief Creates a flatbuffer message, verifies the message and extracts
+  /// metadata.
   ///
-  /// \param BufferPtr Pointer to memory location containing flatbuffer.
-  /// \param Size Size of flatbuffer in bytes.
+  ///\note This constructor should be used as little as possible and then only
```
mattclarke:
```suggestion
  /// \note This constructor should be used as little as possible and then only
```

src/FlatbufferMessage.h
```diff
@@ -9,56 +9,90 @@
...
+  /// in unit tests.
+  /// \param BufferPtr Pointer to memory containing the data.
```
mattclarke: new line before \param

src/FlatbufferMessage.h
```diff
@@ -9,56 +9,90 @@
...
   FlatbufferMessage() = default;
 
-  /// \brief Verifies the data in the flatbuffer to make sure if it is valid.
+  /// \brief Creates a flatbuffer message, verifies the message and extracts
+  /// metadata.
   ///
-  /// \param BufferPtr Pointer to memory location containing flatbuffer.
-  /// \param Size Size of flatbuffer in bytes.
+  ///\note This constructor should be used as little as possible and then only
+  /// in unit tests.
```
SkyToGround:
```suggestion
  /// in unit tests.
  ///
```

src/FlatbufferMessage.h
```diff
@@ -9,56 +9,91 @@
...
+  /// metadata.
+  ///
+  /// \param KafkaMessage The Kafka message used to create the Flatbuffer
+  /// message. \note Will make a copy of the data in the Kafka message.
+  explicit FlatbufferMessage(FileWriter::Msg const &KafkaMessage);
+
+  /// \brief Creates a flatbuffer message, verifies the message and extracts
+  /// metadata.
+  ///
+  /// \param KafkaMessage The copy constructor.
```
mattclarke: This doxygen comment seems to be completely wrong.

src/Stream/MessageWriter.h
```diff
@@ -0,0 +1,54 @@
...
+  void writeMsgImpl(WriterModule::Base *ModulePtr,
+                    FileWriter::FlatbufferMessage const &Msg);
+
+  SharedLogger Log{getLogger()};
+  Metrics::Metric WritesDone{"writes_done",
+                             "Number of completed writes to HDF file."};
+  Metrics::Metric WriteErrors{"write_errors",
+                              "Number of failed HDF file writes.",
+                              Metrics::Severity::ERROR};
+  std::map<ModuleHash, std::unique_ptr<Metrics::Metric>> ModuleErrorCounters;
```
mattclarke: I think these three items can either be public or have a public accessor so the tests can get to these counters without the 'using' workaround.
It is fine line, but it is often reasonable to add things like these to the public API to make testing easier.

src/Stream/MessageWriter.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+using ModuleHash = MessageWriter::ModuleHash;
+
+ModuleHash generateSrcHash(std::string const &Source,
+                           std::string const &FlatbufferId) {
+  return std::hash<std::string>{}(Source + FlatbufferId);
+}
+
+static const ModuleHash UnknownModuleHash{
+    generateSrcHash("Unknown source", "Unknown fb.-id")};
```
mattclarke:
```suggestion
    generateSrcHash("Unknown source", "Unknown fb-id")};
```

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,80 @@
...
+
+#pragma once
+
+#include <concurrentqueue/concurrentqueue.h>
+#include <functional>
+#include <future>
+#include <memory>
+#include <thread>
+
+/// \brief Class for executing jobs in a seperate "worker thread".
```
mattclarke:
```suggestion
/// \brief Class for executing jobs in a separate "worker thread".
```

mattclarke:
> Please leave your thoughts about this. It is a relatively simple matter to replace it with a regular loop if you think so.

I think this should be a simple loop as you suggest. The tests are complicated because of the ThreadedExecutor throwing the timing off. The tests should use a fake ThreadExecutor that runs the tasks immediately then the tests become simple. This fake can be dependency injected into MessageWriter and then we don't need a MessageWriter stand-in.

I think we should try to avoid mocking/faking the class under test where possible as it is a bit of a code smell.

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      SendWork([=]() { RunThread = false; });
+    }
+    WorkerThread.join();
+  }
+  /// \brief Put tasks in the high priority queue.
+  ///
+  /// \param Task The std::function that will be executed when processing the
+  /// task.
+  void SendWork(JobType Task) { TaskQueue.enqueue(std::move(Task)); }
```
mattclarke:
```suggestion
  void sendWork(JobType Task) { TaskQueue.enqueue(std::move(Task)); }
```

src/ThreadedExecutor.h
```diff
@@ -0,0 +1,81 @@
...
+
+  /// \brief Put tasks in the low priority queue.
+  ///
+  /// \param Task The std::function that will be executed when processing the
+  /// task.
+  void SendLowPriorityWork(JobType Task) {
```
mattclarke:
```suggestion
  void sendLowPriorityWork(JobType Task) {
```

mattclarke:
Oh and one other thing: ThreadedExecutor should probably have some of its own tests.

src/Stream/MessageWriter.h
```diff
@@ -0,0 +1,54 @@
...
+  Metrics::Registrar Registrar;
+  static bool const LowPriorityExecutorExit{true};
```
mattclarke:
```suggestion
static constexpr bool LowPriorityExecutorExit{true};
```
I think this would mean that you don't need the `// Must be last` comment below.

src/Stream/MessageWriter.h
```diff
@@ -0,0 +1,54 @@
...
+  ThreadedExecutor Executor{
+      MessageWriter::LowPriorityExecutorExit}; // Must be last
```
mattclarke: Should add why it should be last

src/Stream/MessageWriter.h
SkyToGround: These member variables should really not be public as it would then be way too easy to mess upp the metrics and using public accessors adds otherwise unnecessary code that would only be used by the unit tests (for now at least). Seeing as you say it is a fine line, do you want to have the public accessors?

src/FlatbufferMessage.h
SkyToGround: Oopsie, fixes.

src/Stream/MessageWriter.h
mattclarke: If it is a choice between public accessors or messing about with inheritance and `using` I will (reluctantly) take the accessors.
The other (better) option would be to get the results through the mock registrar, if that was possible?

mattclarke:
> I think this should be a simple loop as you suggest.

I misunderstood the question, I think the ThreadExecutor is fine. And I am an idiot!

src/Stream/MessageWriter.h
SkyToGround: Getting them through the metrics registrar is definitely possible but that would be even more code that would then only be used for these unit tests. Probably on the order 50 to 100 lines for implementing the possibility of doing this. In short, I REALLY do not think it is worth that effort right now.

SkyToGround:
@mattclarke The way the unit tests are designed, there are no timing issues as far as I can tell at least. I know where the suggestion for using an alternative executor for testing comes from as I have seen that talk as well but I would argue that doing that in this particular case would add code and complexity to the tests for no real gain. Also, there is actually no mocking of `Stream::MessageWriter` going on for these unit test.

All the other comments should be addressed now.

mattclarke:
> The way the unit tests are designed, there are no timing issues as far as I can tell at least.

No, there are no timing issues because you use the executor to the run the expectations which is clever, but perhaps too clever?

> I know where the suggestion for using an alternative executor for testing comes from as I have seen that talk as well

I don't know what talk you are referring to, but dependency injection is a well known practice.

> Also, there is actually no mocking of Stream::MessageWriter going on for these unit test.

Okay, perhaps a bit too general. 
Inheriting from a class just to make the protected properties/methods public for testing is a design smell IMO.

Anyway, we've discussed it now and you've thought about it and decided against it, so let's move on.
PR APPROVED!

PR #525, opened 2020-03-18
Added meta data free functions.

SkyToGround:
### Description of work

While waiting for the other PR I thought I could add another one that is pretty much stand alone. This is that.

Having the metadata queries done through free functions makes the code easier to read and understand. Other than that there is not much to say about the implementation. I did have to jump through a few hoops to get unit testing to work and I do not think it was worth the effort in the end.


src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+      KafkaW::getTopicListImpl<UsedProducerMockAlt>("Some_broker", 10ms);
+  EXPECT_EQ(ReceivedTopics, ExpectedTopics);
+}
+
+TEST_F(GetTopicNamesTest, MetadataFailure) {
+  UsedProducerMockAlt::ReturnErrorCode = RdKafka::ErrorCode::ERR__TIMED_OUT;
+  EXPECT_THROW(
+      KafkaW::getTopicListImpl<UsedProducerMockAlt>("Some_broker", 10ms),
+      MetadataException);
+}
```
SkyToGround:
```suggestion
}

```

SkyToGround:
1. To keep the PR small and to avoid breaking the code (i.e. the system tests). However, if you prefer, feel free to review #511 instead. 😄 I will have separate PRs for removing old code when the time comes.
2. I agree that updating mocks is tedious and boring. I would argue that it is not very hard and time consuming in the grand scheme of things though. The reason we want to mock instead of just creating fakes is at least two fold:
   1. Guarantee that we are actually mimicking the interface we are using. I am actually overriding one of the member functions of `Cons`.
   2. Enabling unit testing of some parts of the code that can not be tested otherwise.

With that said, I noticed when looking at the code just now that inheritance from a mocking class  was entirely pointless in at least one case. I will remove that.

matthew-d-jones:
Maybe a good idea to leave the unusedFunction cppcheck warnings in to reduce chance of deadcode later, but there are a couple of `unmatchedSuppression` warnings in `ProducerTests.cpp` which should be sorted out.

SkyToGround:
@matthew-d-jones I will ignore unmatched suppression and unused function warnings from cpp-check for now as they are generally caused by bugs in cpp-check. I will sort it out after we are done merging the other PRs.

src/KafkaW/MetaDataQuery.h
```diff
@@ -0,0 +1,59 @@
...
+#include <string>
+
+namespace KafkaW {
+
+using time_point = std::chrono::system_clock::time_point;
+using duration = std::chrono::system_clock::duration;
+
+// \brief Query the broker for topic + partition offset of a given timestamp.
+//
+// \param[in] Broker    Address of one or several brokers.
```
mattclarke: No `[in]`s please

src/KafkaW/MetaDataQuery.h
```diff
@@ -0,0 +1,59 @@
...
+// \param[in] Topic     Name of topic.
+// \param[in] Partition Partition id.
+// \param[in] Time      Timestamp for which the offset should be found.
+// \param[in] TimeOut   The amount of time to wait for the meta-data call to
```
mattclarke: Table style alignment of descriptions is nice, but inconsistent with the rest of the code base

src/KafkaW/MetaDataQuery.h
```diff
@@ -0,0 +1,59 @@
...
+                 duration TimeOut);
+
+// \brief Query the broker for partitions available on a specified topic.
+//
+// \param[in] Broker    Address of one or several brokers.
+// \param[in] Topic     Name of topic.
+// \param[in] TimeOut   The amount of time to wait for the meta-data call to
+// complete.
+// \return              A set of partition id:s.
+// \throw               MetadataException if a failure occurs.
```
mattclarke: Don't add throws, the chances of this comment being updated if the code changes is almost 0.

src/KafkaW/MetaDataQuery.h
```diff
@@ -0,0 +1,59 @@
...
+                 std::vector<int> const &Partitions, time_point Time,
+                 duration TimeOut);
+
+// \brief Query the broker for partitions available on a specified topic.
+//
+// \param[in] Broker    Address of one or several brokers.
+// \param[in] Topic     Name of topic.
+// \param[in] TimeOut   The amount of time to wait for the meta-data call to
+// complete.
+// \return              A set of partition id:s.
```
mattclarke:
```suggestion
// \return              A set of partition ids.
```

src/KafkaW/MetaDataQueryImpl.cpp
```diff
@@ -0,0 +1,28 @@
...
+findKafkaTopic(const std::string &Topic,
+               const RdKafka::Metadata *KafkaMetadata) {
+  auto Topics = KafkaMetadata->topics();
+  auto Iterator =
+      std::find_if(Topics->cbegin(), Topics->cend(),
+                   [Topic](const RdKafka::TopicMetadata *TopicMetadata) {
+                     return TopicMetadata->topic() == Topic;
+                   });
+  if (Iterator == Topics->end()) {
+    throw MetadataException("Topic \"" + Topic + "\" not listed by broker.");
```
mattclarke: Don't we use fmt to avoid this sort of thing?

src/KafkaW/MetaDataQueryImpl.h
```diff
@@ -0,0 +1,131 @@
...
+    throw MetadataException(
+        "Failed to query broker for available partitions. Error code was: " +
+        std::to_string(ReturnCode));
+  }
+  auto TopicMetaData = findKafkaTopic(Topic, MetadataPtr);
+  std::vector<int> ReturnVector;
+  for (auto const &Partition : *TopicMetaData->partitions()) {
+    ReturnVector.push_back(Partition->id());
+  }
+  delete MetadataPtr;
```
mattclarke: The six lines of code above have complicated lifetimes, I would suggest trying to move 100 to 103 into `findKafkaTopic` and just return the partition vector from there.
Then rename `findKafkaTopic` to reflect this.

MetadataPtr can be a unique_ptr?

src/KafkaW/MetaDataQueryImpl.h
```diff
@@ -0,0 +1,131 @@
...
+    throw MetadataException(
+        "Failed to query broker for available partitions. Error code was: " +
+        std::to_string(ReturnCode));
+  }
+  auto Topics = MetadataPtr->topics();
+  std::set<std::string> TopicNames;
+  for (auto const &CTopic : *Topics) {
+    TopicNames.emplace(CTopic->topic());
+  }
+  delete MetadataPtr;
```
mattclarke: MetadataPtr can be a unique_ptr?

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+public:
+  static int CallsToCreate;
+  static RdKafka::Producer *create(RdKafka::Conf *, std::string &) {
+    UsedProducerMock::CallsToCreate++;
+    return nullptr;
+  }
+};
+int UsedProducerMock::CallsToCreate = 0;
+
+TEST_F(CreateKafkaHandleTest, FailedToCreateHandle) {
```
mattclarke:
```suggestion
TEST_F(CreateKafkaHandleTest, FailueToCreateHandleThrows) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+//
+// Screaming Udder!                              https://esss.se
+
+#include "KafkaW/MetaDataQueryImpl.h"
+#include "helpers/KafkaWMocks.h"
+#include <gtest/gtest.h>
+
+class CreateKafkaHandleTest : public ::testing::Test {};
+
+TEST_F(CreateKafkaHandleTest, Success) {
```
mattclarke:
```suggestion
TEST_F(CreateKafkaHandleTest, OnSuccess) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+};
+int UsedProducerMock::CallsToCreate = 0;
+
+TEST_F(CreateKafkaHandleTest, FailedToCreateHandle) {
+  UsedProducerMock::CallsToCreate = 0;
+  auto testFunc = [](auto Adr) { // Work around for GTest limitation
+    return KafkaW::getKafkaHandle<UsedProducerMock, RdKafka::Conf>(Adr);
+  };
+  EXPECT_THROW(testFunc("some_broker"), MetadataException);
+  EXPECT_EQ(UsedProducerMock::CallsToCreate, 1);
```
mattclarke: Not sure I see the point of this.

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+                       std::string &) override {
+    UsedConfMock::CallsToSet++;
+    return RdKafka::Conf::CONF_INVALID;
+  }
+};
+int UsedConfMock::CallsToCreate = 0;
+int UsedConfMock::CallsToSet = 0;
+
+TEST_F(CreateKafkaHandleTest, FailedToSetBroker) {
+  UsedProducerMock::CallsToCreate = 0;
```
mattclarke: This should be in the fixture if possible

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+  Conf::ConfResult set(std::string const &, std::string const &,
+                       std::string &) override {
+    UsedConfMock::CallsToSet++;
+    return RdKafka::Conf::CONF_INVALID;
+  }
+};
+int UsedConfMock::CallsToCreate = 0;
+int UsedConfMock::CallsToSet = 0;
+
+TEST_F(CreateKafkaHandleTest, FailedToSetBroker) {
```
mattclarke:
```suggestion
TEST_F(CreateKafkaHandleTest, FailureToSetBrokerThrows) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+  UsedProducerMock::CallsToCreate = 0;
+  UsedConfMock::CallsToCreate = 0;
+  UsedConfMock::CallsToSet = 0;
+  auto testFunc = [](auto Adr) { // Work around for GTest limitation
+    return KafkaW::getKafkaHandle<UsedProducerMock, UsedConfMock>(Adr);
+  };
+  EXPECT_THROW(testFunc("some_broker"), MetadataException);
+  EXPECT_EQ(UsedConfMock::CallsToCreate, 1);
+  EXPECT_EQ(UsedProducerMock::CallsToCreate, 0);
```
mattclarke: As above, I am not sure I see the point of these

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+                              RdKafka::Metadata **MetaDataPtr, int) override {
+    *MetaDataPtr = new UsedMockMetadata;
+    return UsedProducerMockAlt::ReturnErrorCode;
+  }
+};
+int UsedProducerMockAlt::TimeOut;
+RdKafka::ErrorCode UsedProducerMockAlt::ReturnErrorCode =
+    RdKafka::ErrorCode::ERR_NO_ERROR;
+
+TEST_F(GetTopicOffsetTest, Success) {
```
mattclarke:
```suggestion
TEST_F(GetTopicOffsetTest, OnSuccess) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+  std::vector<int> Partitions{4, 5};
+  std::vector<std::pair<int, int64_t>> ExpectedResult{{4, RETURN_TIME_OFFSET},
+                                                      {5, RETURN_TIME_OFFSET}};
+  EXPECT_EQ(KafkaW::getOffsetForTimeImpl<UsedProducerMockAlt>(
+                "Some_broker", "some_topic", Partitions,
+                std::chrono::system_clock::now(), 10ms),
+            ExpectedResult);
+}
+
+TEST_F(GetTopicOffsetTest, Failure) {
```
mattclarke:
```suggestion
TEST_F(GetTopicOffsetTest, OnFailureThrowsAndGetsSetToTimeOut) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+class TopicMockAlt {
+public:
+  static RdKafka::Topic *create(RdKafka::Handle *, const std::string &,
+                                RdKafka::Conf *, std::string &) {
+    return nullptr;
+  }
+};
+
+TEST_F(GetTopicPartitionsTest, Success) {
```
mattclarke:
```suggestion
TEST_F(GetTopicPartitionsTest, OnSuccess) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+  UsedProducerMockAlt::ReturnErrorCode = RdKafka::ErrorCode::ERR_NO_ERROR;
+  auto ExpectedPartitions = std::vector<int>{1, 3, 5};
+  auto ReceivedPartitions =
+      KafkaW::getPartitionsForTopicImpl<UsedProducerMockAlt, TopicMockAlt>(
+          "Some_broker", "some_topic", 10ms);
+  EXPECT_EQ(ReceivedPartitions, ExpectedPartitions);
+}
+
+TEST_F(GetTopicPartitionsTest, MetadataFailure) {
```
mattclarke:
```suggestion
TEST_F(GetTopicPartitionsTest, OnGetPartitionsFailureThrows) {
```

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+  UsedProducerMockAlt::ReturnErrorCode = RdKafka::ErrorCode::ERR__TIMED_OUT;
+  try {
+    KafkaW::getPartitionsForTopicImpl<UsedProducerMockAlt, TopicMockAlt>(
+        "Some_broker", "some_topic", 10ms);
+    FAIL() << "Did not throw";
+  } catch (MetadataException &E) {
```
mattclarke: `EXPECT_THROW(<calling code goes here>, MetadataException)`
Better?

src/tests/MetaDataQueryTests.cpp
```diff
@@ -0,0 +1,207 @@
...
+    SUCCEED();
+  } catch (std::exception &E) {
+    FAIL() << "Got the wrong exception.";
+  }
+}
+
+class GetTopicNamesTest : public ::testing::Test {};
+
+TEST_F(GetTopicNamesTest, Success) {
```
mattclarke:
```suggestion
TEST_F(GetTopicNamesTest, OnSuccess) {
```

src/KafkaW/MetaDataQuery.h
SkyToGround: You can blame clang-format I believe. I do not remember doing it. Feels like a waste of time to change it.

src/KafkaW/MetaDataQueryImpl.cpp
SkyToGround: Yes, at least for log messages where the fmt functionality is  automatically included. In this case I could not be bothered to use the fmt-library as it would not have reduced the number of lines of code and I considered this text easy enough to read anyway.

src/tests/MetaDataQueryTests.cpp
SkyToGround: That is fine. You do not have to. 😃 It is simply the result of limitations in mocking in C++ when using free functions.

src/tests/MetaDataQueryTests.cpp
SkyToGround: Ditto.

src/tests/MetaDataQueryTests.cpp
SkyToGround: Feel free to try yourself but that does not work.

src/KafkaW/MetaDataQueryImpl.h
SkyToGround: About `unique_ptr`: No it can not. `std::unique_ptr::get()` is `const`.

src/KafkaW/MetaDataQueryImpl.h
SkyToGround: Ditto about `unique_ptr`.

src/tests/MetaDataQueryTests.cpp
SkyToGround: You should probably have a discussion with @matthew-d-jones about fixtures. 😃 Though I generally agree with you, in this particular case it would add code. Still, I have moved the lines anyway.

src/KafkaW/MetaDataQuery.h
SkyToGround: Why?

src/KafkaW/MetaDataQuery.h
SkyToGround: I do not really follow your line of reasoning here. Can you provide a reference or example where this is a problem?

mattclarke: Go through the code-base and see how many doxy comments are wrong or out-dated, if it is zero then ignore me.
But it is in our code standards not to use throw: https://confluence.esss.lu.se/pages/viewpage.action?pageId=265299412 ;)

src/KafkaW/MetaDataQueryImpl.h
mattclarke: Cool, fair enough.

src/KafkaW/MetaDataQuery.h
mattclarke: Good enough for government work?

src/tests/MetaDataQueryTests.cpp
mattclarke: Stupid GTEST and its macros!
Anyway putting a second set of `()` round the first parameter fixes it. I think overall it is better than what you have currently.
`EXPECT_THROW(
  (KafkaW::getPartitionsForTopicImpl<UsedProducerMockAlt, TopicMockAlt>(
          "Some_broker", "some_topic", 10ms)),
 MetadataException);`

PR #527, opened 2020-03-26
Move dataset classes into separate directory and object library

matthew-d-jones:
### Description of work

Some of the classes in the `NeXusDataset` namespace were living in writer module directories and the rest roaming the great plains of `src`. Seems sensible to me to move them and their tests into a separate directory and object library.

I've done this in preparation for adding EPICS alarm writing but thought it best to create a separate PR to minimise chance for merge conflicts and keep the alarm PR easy to review.

Improved names of some namespaces, see commit messages.


src/NeXusDataset/ExtensibleDataset.h
```diff
@@ -14,7 +14,7 @@
 
 #pragma once
 
-#include "logger.h"
+#include "../logger.h"
```
SkyToGround: I thought that the _src_ directory was set as an include directory but if it is not, it should be done so as to avoid relative include paths like this one.

matthew-d-jones:
I made the directory name match the namespace name, which is consistent with the other directories except "fb_metadata_extractors". Any reason not to change "fb_metadata_extractors" to "FlatbufferMetadata" to match the namespace?

src/NeXusDataset/ExtensibleDataset.h
matthew-d-jones: I made a conscious decision not to do that because I thought it would encourage people to consider whether they should be adding dependency with other parts of the codebase. But I could be persuaded otherwise...

SkyToGround: I can see your reasoning (though I do not believe it will make much of a difference). My reason for disliking relative paths when including headers is because I feel that it makes it harder to (manually)  find files in a repository and it makes re-factoring of code a bit more tedious. But to be honest it is not that bad if you only have to go back one level (i.e. "../"). It is worse when there are several levels of relative paths (e.g. "../../" or "../some_path/").

matthew-d-jones:
@SkyToGround 

1. No, but I did find an unused "using" statement when I grepped to check.
2. Naming is hard... I don't think the current name is terrible.
3. :+1: done

PR #528, opened 2020-03-27
Use python streaming types library for system tests

matthew-d-jones:
### Description of work

Start using the python-streaming-data-types library in system-tests, for run start and stop.


system-tests/helpers/kafkahelpers.py
```diff
@@ -30,15 +28,15 @@ def publish_run_start_message(
...
+    runstart_message = serialise_pl72(
         job_id,
         nexus_filename,
         start_time,
         stop_time,
         nexus_structure=nexus_structure,
         service_id=service_id,
     )
-    producer.produce(topic, runstart_message)
+    producer.produce(topic, bytes(runstart_message))
```
SkyToGround: It seems to me that either there is a bug in `serialise_pl72` or there is a bug in the Flatbuffers library if use of `bytes` here is required. If it is not, it should be removed, otherwise the bug should be fixed in `serialise_pl72`.

system-tests/helpers/kafkahelpers.py
```diff
@@ -50,10 +48,9 @@ def publish_run_stop_message(
     stop_time: Optional[int] = None,
     service_id: Optional[str] = None,
 ) -> str:
-    runstop_message = create_runstop_message(
-        job_id, service_id=service_id, stop_time=stop_time
-    )
-    producer.produce(topic, runstop_message)
+    runstop_message = serialise_6s4t(job_id, service_id=service_id, stop_time=stop_time)
+
+    producer.produce(topic, bytes(runstop_message))
```
SkyToGround: Ditto about use of `bytes` here as well.

system-tests/helpers/kafkahelpers.py
matthew-d-jones: NICOS uses a different Kafka library which requires `bytearray` rather than `bytes`. So we have to cast here or in NICOS whichever type we choose to output from the library.

PR #530, opened 2020-03-30
Filter refactor part 4 filter on source

SkyToGround:
### Issue

* Several. If you read this you know of the relevant JIRA tickets.

### Description of work

* There are some unused function cppcheck warnings that I still will not fix as they will be resolved properly a little bit down the road.
* Thinking a little bit more about using the `using` statement for accessing protected members, I find that this is really the least ridiculous way of accessing a bunch of counters that really does not (currently) have any business  in being public. So unless a majority of the reviewers thinks that this should change, I will leave it as is.
* Based on previous discussions, I have removed the Doxygen documentation from `class SourceFilter`.
* There is a flow diagram that was the basis for the code implemented in this PR. If requested I can add that to the PR or otherwise make it available.


src/Stream/SourceFilter.cpp
```diff
@@ -0,0 +1,78 @@
...
+  CurrentTimeStamp = InMsg.getTimestamp();
+
+  if (CurrentTimeStamp < toNanoSec(Start)) {
+    if (BufferedMessage.isValid()) {
+      MessagesDiscarded++;
+    }
+    BufferedMessage = InMsg;
+    return false;
+  }
+  if (CurrentTimeStamp > toNanoSec(Start) and
```
mattclarke: Just wondering if that should be `>=`?

mattclarke: Actually, you are already checking for < Start above, could this just be a check for > Stop, and then swap the logic with that below. Okay that is as clear as mud, let's try via code:
```
if (CurrentTimeStamp > toNanoSec(Stop)) {
  IsDone = true;
}

sendBufferedMessage();
sendMessage(InMsg);
return true;


```

SkyToGround: Yes example code is almost always better. Looking at it I feel stupid for not having seen that change myself.

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+TEST_F(SourceFilterTest, SameTSBeforeStart) {
+  REQUIRE_CALL(Writer, addMessage(_)).TIMES(1);
+  auto UnderTest = getTestFilter();
+  UnderTest->addDestinationPtr(0);
+  auto TestMsg = generateMsg();
+  UnderTest->filterMessage(std::move(TestMsg));
+  TestMsg = generateMsg();
+  UnderTest->filterMessage(std::move(TestMsg));
+  EXPECT_TRUE(UnderTest->MessagesReceived == 2);
+  EXPECT_TRUE(UnderTest->RepeatedTimestamp == 1);
```
matthew-d-jones: What is the purpose of tracking repeated timestamps?

src/Stream/SourceFilter.h
```diff
@@ -0,0 +1,74 @@
...
+#include "Metrics/Registrar.h"
+#include "Stream/MessageWriter.h"
+#include <chrono>
+
+namespace Stream {
+
+using time_point = std::chrono::system_clock::time_point;
+uint64_t toNanoSec(time_point Time);
+
+class SourceFilter {
```
matthew-d-jones: Would benefit from a description comment. Particularly to explain the role of the buffered message and connection with the forwarder periodic-update functionality.

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+TEST_F(SourceFilterTest, BufferedMessageBeforeStart) {
+  REQUIRE_CALL(Writer, addMessage(_)).TIMES(1);
+  auto UnderTest = getTestFilter();
+  UnderTest->addDestinationPtr(0);
+  auto TestMsg = generateMsg();
+  UnderTest->filterMessage(std::move(TestMsg));
+  EXPECT_TRUE(UnderTest->MessagesReceived == 1);
+  EXPECT_TRUE(UnderTest->MessagesTransmitted == 0);
+  EXPECT_FALSE(UnderTest->hasFinished());
```
matthew-d-jones: What happens if the system time reaches the stop time plus leeway and this filter has not received another message? Will something else have to tell it to release its buffered message to the writer module at that point? `sendBufferedMesssage()` isn't in the public interface...

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+
+TEST_F(SourceFilterTest, BufferedMessageBeforeStart) {
+  REQUIRE_CALL(Writer, addMessage(_)).TIMES(1);
+  auto UnderTest = getTestFilter();
+  UnderTest->addDestinationPtr(0);
+  auto TestMsg = generateMsg();
+  UnderTest->filterMessage(std::move(TestMsg));
+  EXPECT_TRUE(UnderTest->MessagesReceived == 1);
+  EXPECT_TRUE(UnderTest->MessagesTransmitted == 0);
```
matthew-d-jones: This is a good opportunity to documented why you cache the message and don't transmit it yet.

src/tests/Stream/SourceFilterTest.cpp
SkyToGround: See destructor.

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+  EXPECT_FALSE(UnderTest->hasFinished());
+}
+
+TEST_F(SourceFilterTest, MultipleMessagesBeforeStart) {
```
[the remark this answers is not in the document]
SkyToGround: Not really, it executes code that is not executed elsewhere. Specifically:
```
if (BufferedMessage.isValid()) {
  MessagesDiscarded++;
}
```

src/tests/Stream/SourceFilterTest.cpp
SkyToGround: See DM-1869. I think there is another JIRA ticket for this as well.

src/tests/Stream/SourceFilterTest.cpp
SkyToGround: Done.

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+  EXPECT_TRUE(UnderTest->MessagesDiscarded == 1);
+  EXPECT_FALSE(UnderTest->hasFinished());
+}
+
+TEST_F(SourceFilterTest, MessageNoDest) {
+  FORBID_CALL(Writer, addMessage(_));
+  auto UnderTest = getTestFilter();
+  auto TestMsg = generateMsg();
+  UnderTest->filterMessage(std::move(TestMsg));
+  EXPECT_FALSE(UnderTest->hasFinished());
```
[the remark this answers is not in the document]
SkyToGround: Considering that the first line of the test says: `FORBID_CALL(Writer, addMessage(_));`. I do not think this is a problem.

src/tests/Stream/SourceFilterTest.cpp
```diff
@@ -0,0 +1,282 @@
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This code has been produced by the European Spallation Source
+// and its partner institutes under the BSD 2 Clause License.
+//
+// See LICENSE.md at the top level for license information.
+//
+// Screaming Udder!                              https://esss.se
+
+/// \brief Test partition filtering.
```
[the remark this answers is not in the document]
SkyToGround: Removed.

src/Stream/SourceFilter.h
```diff
@@ -0,0 +1,74 @@
...
+  SourceFilter(time_point StartTime, time_point StopTime,
+               MessageWriter *Destination, Metrics::Registrar RegisterMetric);
+  virtual ~SourceFilter();
+  void addDestinationPtr(Message::DestPtrType NewDestination) {
+    DestIDs.push_back(NewDestination);
+  };
+
+  virtual bool filterMessage(FileWriter::FlatbufferMessage &&InMsg);
+  void setStopTime(time_point StopTime);
+  time_point getStopTime() const { return Stop; }
```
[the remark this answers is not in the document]
SkyToGround: I am aware. The function is used by code that will be added through an upcoming PR.

src/Stream/SourceFilter.h
SkyToGround: I have added a short description. You will have to discuss the matter with @mattclarke if it is not to your satisfaction.

src/tests/Stream/SourceFilterTest.cpp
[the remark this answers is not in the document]
matthew-d-jones: I saw the `FORBID_CALL`, I mean it's not immediately obvious why you want to enforce that the filter reports not to have finished in this situation. Are you intentionally making sure the filter can't report finished until it has had a destination set, so that it can't finish before it has been set up properly or something?

src/Stream/SourceFilter.h
matthew-d-jones: I think putting the diagram on confluence and adding a link in the comment is a good idea. I also think adding _something_ about the role of `BufferedMessage` is a good idea. Maybe something like this? "SourceFilter buffers a message such that, when used in conjunction with the periodic-update feature of the forwarder, the writer module should always be able to record at least the data from a single message." At least the reader would have some clue what to ask colleagues about to understand what is going on here.

src/Stream/SourceFilter.h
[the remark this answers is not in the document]
matthew-d-jones: I think you're pulling these PRs out of another branch which maybe contains the code that uses it? If that is the case I guess the risk of this remaining dead code is low. I think we should try to avoid temporary dead code like this where possible though, particularly when cppcheck isn't providing a safety net for it, for when we merge v3 back to master.

src/tests/Stream/SourceFilterTest.cpp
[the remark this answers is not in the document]
SkyToGround: Thats fair, the last line is/was somewhat redundant here.

src/Stream/SourceFilter.h
SkyToGround: If you think it is good to have the diagram on confluence then I think it is better to have it be part of the repository. It is hard to find things there as it is, especially if you do not have access. As for the text, come up with something that both you and @mattclarke agrees on and we an add that.

matthew-d-jones: :+1: fine with me, I only suggested confluence because I thought it was dictated that that is where we must put the documentation. I'll talk with Matt about the text.

SkyToGround:
Seeing as you have both accepted the PR. I will merge it without further changes.

PR #532, opened 2020-04-01
Rename StreamMaster to StreamController

mattclarke:
### Issue

BCG-246 (Oops!)

### Description of work

Renamed class and associated classes and variables so it matches the name on the design documents in Confluence.

### DIFF
--- a/src/CommandParser.h
+++ b/src/CommandParser.h
@@ -7,10 +7,6 @@
 //
 // Screaming Udder!                              https://esss.se
 
-/// \file This file defines the different success and failure status that the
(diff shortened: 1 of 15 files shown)


src/CommandParser.h
```diff
@@ -7,10 +7,6 @@
 //
 // Screaming Udder!                              https://esss.se
 
-/// \file This file defines the different success and failure status that the
-/// `StreamMaster` and the `Streamer` can incur. These error object have some
-/// utility methods that can be used to test the more common situations.
-
```
matthew-d-jones, line 13: This comment appears to be copy-pasted from StreamStatus and doesn't belong here at all, so I deleted it.

PR #534, opened 2020-04-06
Speed up system tests

SkyToGround:
### Description of work

Maybe this was a bit of wast of time. On the other hand, I know some other people that I feel are wasting a bunch of time right now. 😃 
In the end, this PR seems to remove ~10 minutes from running the system tests on the first run and likely a few more minutes on subsequent runs on the same build node.

@matthew-d-jones I changed back to using Ubuntu as a base image (away from Alpine) due to Ubuntu not requiring us to build _boost_build_ and _flatbuffers_.


system-tests/create_filewriter_image.py
```diff
@@ -0,0 +1,184 @@
...
+        container = generate_new_container()
+    copy_to_container(container)
+    try:
+        container.start()
+    except docker.errors.APIError as e:
+        pass
+
+    if conan_hash_changed(container) or src_hash_changed(container):
+        container.exec_run('bash -c "rm -rf *"', workdir="/home/jenkins/build/")
+        run_conan(container)
```
matthew-d-jones, line 175: Does the conan bit need to be rerun if only the src hash has changed?

system-tests/create_filewriter_image.py
```diff
@@ -0,0 +1,184 @@
...
+        if var_name in os.environ:
+            environment_variables[var_name] = os.environ[var_name]
+
+    add_var("http_proxy")
+    add_var("https_proxy")
+    add_var("local_conan_server")
+    try:
+        client.images.get(IMAGE_NAME)
+    except docker.errors.ImageNotFound as e:
+        client.images.pull(IMAGE_NAME)
```
matthew-d-jones, line 76: Pulling the image can take a while, a print statement here would be good.

system-tests/create_filewriter_image.py
```diff
@@ -0,0 +1,184 @@
...
+    container = client.containers.create(
+        IMAGE_NAME,
+        name=CONTAINER_NAME,
+        command="tail -f /dev/null",
+        environment=environment_variables,
+    )
+    container.start()
+    container.exec_run("apt-get --assume-yes install kafkacat", user="root")
```
matthew-d-jones, line 84: Maybe also a print here? I found there was a lot of silence after seeing "Unable to find container, re-generating"

system-tests/create_filewriter_image.py
SkyToGround: In principle no. However, because the same (base) container is used for every system regardless of which PR, I am worried that the container will get into a state where it is not possible to build the file writer without wiping the CMake cache files. Hence the need to re-run Conan in order to re-generate the relevant supporting files.

system-tests/create_filewriter_image.py
```diff
@@ -0,0 +1,184 @@
...
+        pass
+
+
+def run_conan(container):
+    print("Running conan")
+    container.exec_run("mkdir build", workdir="/home/jenkins/")
+    execute_command(
+        "conan install --build=outdated ../conan", "/home/jenkins/build", container
+    )
+    execute_command("cp conan_hash_new.txt conan_hash.txt", "/home/jenkins/", container)
```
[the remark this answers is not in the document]
SkyToGround: Fixed.

system-tests/create_filewriter_image.py
SkyToGround: 👍

system-tests/create_filewriter_image.py
SkyToGround: Done

SkyToGround:
@matthew-d-jones Yes, the conan cache will grow in the container used as a base. As the code deletes the build directory if anything has changed in the source code, I think we can get away with not explicitly forcing a completely clean re-build of the image. Furthermore, according to Afonso, the images are actually cleaned out every day nowdays (at around 4 in the moring) which should reduce the probability of something breaking somewhat.

SkyToGround:
I have added code for generating a "cleaned" docker image and some documentation explaining this fact.

PR #535, opened 2020-04-07
Filter refactor part 5 edits

matthew-d-jones:
Some improvements to tests for the branch, mostly by @mattclarke .
Could do with more dependency injection to avoid having to subclass for testing, I've created ticket DM-1900 for this.


src/Msg.h
```diff
@@ -64,7 +64,7 @@ struct Msg {
     }
     return Size;
   }
-  MessageMetaData const getMetaData() const { return MetaData; }
+  MessageMetaData getMetaData() const { return MetaData; }
```
SkyToGround, line 67: I made this function have a `const` return value due to a bug I introduced in my code and to make it impossible to re-introduce that bug. I see no reason to change that back.
```suggestion
  MessageMetaData const getMetaData() const { return MetaData; }
```

src/Stream/Partition.cpp
```diff
@@ -82,26 +84,22 @@ bool Partition::shouldStopBasedOnPollStatus(Kafka::PollStatus CStatus) {
 void Partition::pollForMessage() {
   auto Msg = ConsumerPtr->poll();
   switch (Msg.first) {
-  case Kafka::PollStatus::Message:
```
SkyToGround, line 85: I consider this change a bug. There are other metrics counters for keeping track of number of processed messages. I have created the relevant (code) "suggestions" for moving the messages received counter back to where it should be.

src/Stream/Partition.cpp
```diff
@@ -82,26 +84,22 @@ bool Partition::shouldStopBasedOnPollStatus(Kafka::PollStatus CStatus) {
 void Partition::pollForMessage() {
   auto Msg = ConsumerPtr->poll();
   switch (Msg.first) {
```
SkyToGround, line 86:
```suggestion
  switch (Msg.first) {
      case Kafka::PollStatus::Message:
      MessagesReceived++;		
     break;
```

src/Stream/Partition.cpp
```diff
@@ -115,6 +113,7 @@ void Partition::pollForMessage() {
 }
 
 void Partition::processMessage(FileWriter::Msg const &Message) {
+  MessagesReceived++;
```
SkyToGround, line 116:
```suggestion
```

src/Stream/Partition.cpp
```diff
@@ -82,26 +84,22 @@ bool Partition::shouldStopBasedOnPollStatus(Kafka::PollStatus CStatus) {
...
   case Kafka::PollStatus::TimedOut:
     KafkaTimeouts++;
     break;
   case Kafka::PollStatus::Error:
     KafkaErrors++;
     break;
-  case Kafka::PollStatus::EndOfPartition: // Do nothing
-    break;
-  case Kafka::PollStatus::Empty: // Do nothing
+  default:
```
SkyToGround, line 93: I think (though I am not sure) that _clang-tidy_ complains if not all enum possibilities are accounted for in a switch statement. I am fine with leaving this as is though.

src/Msg.h
matthew-d-jones: Ok, I changed it because it didn't look useful and clang-tidy labels it as a probable mistake. Adding a comment to explain why it is there would help avoid someone making the same change in the future.

src/Stream/Partition.cpp
matthew-d-jones: I don't think it is too important if it is changed back, but I'd like to understand your comment if possible.
"There are other metrics counters for keeping track of number of processed messages", it isn't being used for that. The changed code looks functionally equivalent to me, it just avoids doing the comparison `Msg.first == Kafka::PollStatus::Message` twice. The only risk I see is that someone could now put some code between the poll and the call to `processMessage` which calls `return` and thus introduce a bug of the message not being counted.

src/Stream/Partition.cpp
matthew-d-jones: clang-tidy isn't complaining at me. If `default` wasn't there the compiler would warn (and thus error).

src/Stream/Partition.cpp
matthew-d-jones: Moving `MessagesReceived++;` to after the `if (Msg.first == Kafka::PollStatus::Message) {` but outside of `processMessage()` would probably be better.

src/Msg.h
SkyToGround: I have added a comment to the suggestion.

src/Stream/Partition.cpp
SkyToGround: Actually, saying "bug" in this case was not entirely correct. Still, my reasoning is that `MessagesReceived` indicated just that. In principle `shouldStopBasedOnPollStatus()` could return `false` (though it does not in the current implementation) which would then result in this metric being wrong. Furthermore, why move `MessagesReceived++;` it at all when the switch statement is used for the other counters as well? As for the extra comparison, without evidence I do not believe this to impact the performance in any measurable way.

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
   REQUIRE_CALL(*Consumer, poll()).TIMES(1).LR_RETURN(std::move(PollReturn));
-  FORBID_CALL(*UnderTest, processMessage(_));
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 0);
```
SkyToGround, line 132:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 0);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
   REQUIRE_CALL(*Consumer, poll()).TIMES(1).LR_RETURN(std::move(PollReturn));
-  REQUIRE_CALL(*UnderTest, processMessage(_)).TIMES(1);
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 1);
```
SkyToGround, line 141:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 1);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
-  FORBID_CALL(*UnderTest, processMessage(_));
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 0);
+  EXPECT_EQ(UnderTest->KafkaTimeouts.getCounterPtr()->load(), 1);
```
SkyToGround, line 151:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 0);
  EXPECT_EQ(int(UnderTest->KafkaTimeouts), 1);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
-  FORBID_CALL(*UnderTest, processMessage(_));
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 0);
+  EXPECT_EQ(UnderTest->KafkaErrors.getCounterPtr()->load(), 1);
```
SkyToGround, line 161:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 0);
  EXPECT_EQ(int(UnderTest->KafkaErrors), 1);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
   REQUIRE_CALL(*Consumer, poll()).TIMES(1).LR_RETURN(std::move(PollReturn));
-  FORBID_CALL(*UnderTest, processMessage(_));
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 0);
```
SkyToGround, line 170:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 0);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
-  REQUIRE_CALL(*UnderTest, processMessage(_)).TIMES(1);
-  REQUIRE_CALL(*UnderTest, addPollTask()).TIMES(1);
-  REQUIRE_CALL(*UnderTest, shouldStopBasedOnPollStatus(_))
-      .TIMES(1)
-      .RETURN(false);
-  UnderTest->pollForMessageBase();
-  EXPECT_FALSE(UnderTest->hasFinished());
+  UnderTest->pollForMessage();
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 1);
+  EXPECT_EQ(UnderTest->FlatbufferErrors.getCounterPtr()->load(), 1);
```
SkyToGround, line 195:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 1);
  EXPECT_EQ(int(UnderTest->FlatbufferErrors), 1);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -107,259 +108,156 @@ class PartitionTest : public ::testing::Test {
...
   UnderTest->MsgFilters.clear();
   size_t SomeOtherHash{42};
   UnderTest->MsgFilters[SomeOtherHash] = std::move(TestFilter);
-  FORBID_CALL(*TestFilterPtr, filterMessage(_));
   setExtractorModule<zzzzFbReader>("zzzz");
   FileWriter::Msg Msg(SomeData.data(), SomeData.size());
-  UnderTest->processMessageBase(Msg);
+  UnderTest->processMessage(Msg);
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 1);
+  EXPECT_EQ(UnderTest->MessagesProcessed.getCounterPtr()->load(), 0);
```
SkyToGround, line 257:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 1);
  EXPECT_EQ(int(UnderTest->MessagesProcessed), 0);
```

src/tests/Stream/PartitionTests.cpp
```diff
@@ -368,10 +266,12 @@ TEST_F(PartitionTest, ProcessMessageSourceHashIsKnown) {
   REQUIRE_CALL(*TestFilterPtr, hasFinished()).TIMES(1).RETURN(false);
   setExtractorModule<zzzzFbReader>("zzzz");
   FileWriter::Msg Msg(SomeData.data(), SomeData.size());
-  UnderTest->processMessageBase(Msg);
+  UnderTest->processMessage(Msg);
+  EXPECT_EQ(UnderTest->MessagesReceived.getCounterPtr()->load(), 1);
+  EXPECT_EQ(UnderTest->MessagesProcessed.getCounterPtr()->load(), 1);
```
SkyToGround, line 271:
```suggestion
  EXPECT_EQ(int(UnderTest->MessagesReceived), 1);
  EXPECT_EQ(int(UnderTest->MessagesProcessed), 1);
```

src/Msg.h
mattclarke: Could be const-ref?
Everywhere it is used (currently) is to access a value inside MetaData, so lifetime is not a concern.

src/Stream/Partition.cpp
mattclarke: Done!

src/Stream/Partition.cpp
mattclarke: Done!

src/Msg.h
SkyToGround: Sure, I do not see a problem with doing that.

matthew-d-jones:
@SkyToGround @mattclarke Is everyone happy that we merge this and then Jonas's branch now?

SkyToGround:
@matthew-d-jones Now I am.

matthew-d-jones:
oops sorry, didn't notice we didn't revert that yet.

PR #536, opened 2020-04-07
Activate metrics code.

SkyToGround:
### Description of work

As the title says. A relatively simple PR.


src/MainOpt.h
```diff
@@ -48,6 +48,9 @@ struct MainOpt {
   /// Can optionally use the `graylog_logger` library to log to this address.
   uri::URI GraylogLoggerAddress;
 
+  uri::URI GrafanCarbonAddress;
```
matthew-d-jones, line 51: Typo, "GrafanCarbonAddress" -> "GrafanaCarbonAddress". But maybe you meant "GraphiteCarbonAddress" anyway? I would suggest just "CarbonAddress".

src/CLIOptions.cpp
```diff
@@ -145,6 +145,11 @@ void setCLIOptions(CLI::App &App, MainOpt &MainOptions) {
   addUriOption(App, "--graylog-logger-address",
                MainOptions.GraylogLoggerAddress,
                "<host:port> Log to Graylog via graylog_logger library");
+  addUriOption(App, "--grafana-carbon-address", MainOptions.GrafanCarbonAddress,
+               "<host:port> Address to the Grafana (Carbon) metrics service.");
+  App.add_option("--metrics-id", MainOptions.GrafanaCarbonMetricsId,
+                 "An extra metrics ID string. Only used if present. Will take "
+                 "the form: \"kafka-to-nexus.[id].*\"");
```
matthew-d-jones, line 152: Tobias has requested reintroduction of a configurable "service_id" in status messages. Perhaps we should add that as an option instead (and always use it as part of the metrics name if it is set)?

src/MainOpt.h
SkyToGround: The intent was `GrafanaCarbonAddress`. I suspect that most people using Grafana does not know what Graphite is and that even fewer knows what Carbon is. Hence picking that name instead of "CarbonAddress".

src/CLIOptions.cpp
SkyToGround: Sure.

SkyToGround:
@matthew-d-jones I believe that I have addressed your comments now.

src/MainOpt.h
matthew-d-jones: If they know about Grafana and not Carbon won't this just lead to them putting the wrong address in the field? I'm a little uneasy about linking it to Grafana when one or more facilities may end up using Grafana with a different database, or Carbon with a different dash.
PSI were using InfluxDB, I don't know if that is still the case.

SkyToGround: Based on a comment from Mark last week, I believe that they do not use Grafana at all. Now, your comment is fair but I do not agree with it. So: @mattclarke, can we get a tiebreaker here?

PR #537, opened 2020-04-13
Filter refactor part 6, Topic logic

SkyToGround:
### Description of work

* Added the code for the Kafka topic logic
* Added a helper function to the Kafka code and some settings to `struct BrokerSettings`.
* Reverted use of `std::unique_ptr` in `Stream::Partition` as it caused a race condition.
* Disabled `Stream::Partition` unit tests.
* Please talk to me before making any changes to the `Stream::Topic` interface or better yet, make the relevant corresponding changes to the *DM-1803_part1* branch.
* We are now down to two false positives from cppcheck. These two will also be gone with the next (and last) PR.


src/Stream/Topic.h
```diff
@@ -0,0 +1,77 @@
...
+                           duration TimeOut) const;
+
+  virtual std::vector<int>
+  getPartitionsForTopicInternal(std::string Broker, std::string Topic,
+                                duration TimeOut) const;
+
+  std::vector<std::unique_ptr<Partition>> ConsumerThreads;
+  ThreadedExecutor Executor; // Must be last
+};
+} // namespace Stream
```
SkyToGround, line 77:
```suggestion
} // namespace Stream

```

SkyToGround:
There are several good reasons not to do the metadata queries elsewhere:
* Limit responsibility (make the code easier to understand)
* Have the metadata requests be multithreaded (this is one of the reasons for this re-write)
* Avoid excessive number of arguments

As far as I can see, there are three alternative implementations to the one we are using here:
* Make `Topic` a templated class and do dependency injection using template arguments. This is ugly and you have argued against this before on compilation time grounds and I lean towards agreeing with that.
* Use the ASIO library instead of `Executor` for executing the metadata calls. This option is bad as it means replacing some hard to understand code with some code that is even harder to understand. It is really not an option in my opinion.
* Use a `while`-loop instead of `Executor`. This is not "flashy" but it is easy to understand and relatively easy to write and test. Overrides would probably still be required to test the `while` loop but to a lesser extent. This is what I should have done to from the start. Re-writing the code to use a while loop will probably be required in order to implement some features going forward as well.

SkyToGround:
Oh, and I am pretty sure that we are already at the minimum number of metadata calls already so doing these higher up in the hierarchy would not reduce the number of them anyway.

matthew-d-jones:
Here you do one metadata call and one offset query per topic right?
I'm suggesting one metadata call and one offset query to cover all topics. As little could be done until this one query happened there wouldn't be an advantage to doing it asynchronously. `Topic` would then look up the `TopicPartitions` it is interested in from the big packet of all metadata. Passing a single struct of metadata down to `Topic` seems like a simple approach to me, given that it would avoid the concurrency altogether. /shrug, just an idea.

SkyToGround:
I had to double check the Kafka documentation and I think you are probably right on that one. It looks like it is possible to get all partition IDs when you get the list of topics. Still, we really do want the concurrency as slow metadata calls (including offset queries) is one of the major issues that we want to solve with this re-factor. There are solutions other than this one for doing this that could be better but I suggest that we look into it again when we are done with merging v3 into master.

matthew-d-jones:
I guess it is putting all eggs in one basket, having one request, but I think the situation before was to do a load of queries in series or in parallel but held up by the slowest? Anyway, agree we proceed with your impl for time being. Sorry, I'd forgotten to add reviewers to #545, but hopefully you saw it in the previous comments.

SkyToGround:
It builds, so I am merging now.

PR #545, opened 2020-05-17
Revert dependency injection of executor

matthew-d-jones:
- Reverts `PartitionTests` changes that were made to enable dependency injection of Executor since this turns out to be unviable.

- Use a factory to create consumers so we avoid making real consumers in `TopicTests`. Since poll is not re-called in the tests we were guaranteed to have to wait for the consumer destruction timeout of 5 seconds, which on my machine is the same as the run time for all other unit tests put together.

- Added `const &` for many method parameters in `Topic`.

- Minor refactoring in `TopicTests` to aid understanding of what is being tested and why.

### DIFF
--- a/src/Stream/Topic.cpp
+++ b/src/Stream/Topic.cpp
@@ -15,6 +15,20 @@
 
 namespace Stream {
 
+Topic::Topic(Kafka::BrokerSettings const &Settings, std::string const &Topic,
+             SrcToDst Map, MessageWriter *Writer,
+             Metrics::Registrar &RegisterMetric, time_point StartTime,
+             duration StartTimeLeeway, time_point StopTime,
+             duration StopTimeLeeway,
+             std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers)
+    : KafkaSettings(Settings), TopicName(Topic), DataMap(std::move(Map)),
+      WriterPtr(Writer), StartConsumeTime(StartTime),
+      StartLeeway(StartTimeLeeway), StopConsumeTime(StopTime),
+      StopLeeway(StopTimeLeeway),
+      CurrentMetadataTimeOut(Settings.MinMetadataTimeout),
+      Registrar(RegisterMetric.getNewRegistrar(Topic)),
+      ConsumerCreator(std::move(CreateConsumers)) {}
+
 void Topic::start() {
   Executor.sendWork([=]() { initMetadataCalls(KafkaSettings, TopicName); });
 }
--- a/src/Stream/Topic.cpp
+++ b/src/Stream/Topic.cpp
@@ -20,7 +20,7 @@ Topic::Topic(Kafka::BrokerSettings const &Settings, std::string const &Topic,
              Metrics::Registrar &RegisterMetric, time_point StartTime,
              duration StartTimeLeeway, time_point StopTime,
              duration StopTimeLeeway,
-             std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers)
+             std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers = std::make_unique<Kafka::ConsumerFactory>())
     : KafkaSettings(Settings), TopicName(Topic), DataMap(std::move(Map)),
       WriterPtr(Writer), StartConsumeTime(StartTime),
       StartLeeway(StartTimeLeeway), StopConsumeTime(StopTime),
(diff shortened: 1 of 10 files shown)


src/Stream/Topic.h
```diff
@@ -25,7 +26,15 @@ class Topic {
   Topic(Kafka::BrokerSettings const &Settings, std::string const &Topic,
         SrcToDst Map, MessageWriter *Writer, Metrics::Registrar &RegisterMetric,
         time_point StartTime, duration StartTimeLeeway, time_point StopTime,
-        duration StopTimeLeeway);
+        duration StopTimeLeeway,
+        std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers)
```
SkyToGround, line 30:
```suggestion
        std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers = {})
```
What was it about changing interfaces?

src/tests/Stream/PartitionTests.cpp
```diff
@@ -58,8 +53,8 @@ class PartitionStandIn : public Stream::Partition {
                    Stream::time_point Stop, Stream::duration StopLeeway,
                    Stream::duration KafkaErrorTimeout)
       : Stream::Partition(std::move(Consumer), Partition, std::move(TopicName),
-                          Map, Writer, RegisterMetric, Start, Stop, StopLeeway,
-                          KafkaErrorTimeout) {}
+                          Map, Writer, std::move(RegisterMetric), Start, Stop,
```
SkyToGround, line 56: I am pretty sure that `std::move` is a no-op in this case and thus useless.

src/tests/Stream/PartitionTests.cpp
```diff
@@ -77,9 +72,19 @@ class PartitionStandIn : public Stream::Partition {
...
+  // Queue a job in the executor and block until it is complete
+  // so that we know previously queued job that is part of test should
+  // now have been executed
+  std::promise<bool> Promise;
+  auto Future = Promise.get_future();
+  UnderTest->Executor.sendWork([&Promise]() { Promise.set_value(true); });
+  Future.wait();
+}
+
 // Tests disabled due to being sensitive to thread scheduling
```
SkyToGround, line 85:
```suggestion
```

src/Stream/Topic.h
```diff
@@ -25,7 +26,15 @@ class Topic {
...
+      : KafkaSettings(Settings), TopicName(Topic), DataMap(std::move(Map)),
```
SkyToGround, line 31: I do not see a good reason to putting the constructor implementation in the header file. It is not helpful to the user of this class to have it here I think. Do you have a convincing reference which argues for doing it?

src/Stream/Topic.h
matthew-d-jones: Oops, forgot about that, sorry.
Your suggestion looks a bit risky though? Isn't it better that your branch fails to build so that it is clear you need to update it to pass the dependency? (Or have I misunderstood? I haven't finished my first coffee of the day yet)

src/tests/Stream/PartitionTests.cpp
matthew-d-jones: Doesn't look like it to me, and `clang-tidy` suggested the change and usually gets that one right.

src/Stream/Topic.h
matthew-d-jones: In past projects I was encouraged to put trivial, initialiser-list constructors in the header as it doesn't make a difference to the compiler in this case and means you only have to change it in one place. I don't have strong feelings about it, it is just habit. I'd forgotten you prefer the opposite; I'll put it back.

matthew-d-jones:
@SkyToGround If I've misunderstood the `std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers = {}` thing and you are convinced it is a good idea then feel free to make that change and merge this.

src/tests/Stream/PartitionTests.cpp
matthew-d-jones: done

src/Stream/Topic.h
matthew-d-jones: done

src/Stream/Topic.h
SkyToGround: Oops, I got the types confused. Give me a moment.

src/Stream/Topic.cpp
```diff
@@ -19,20 +19,22 @@ Topic::Topic(Kafka::BrokerSettings const &Settings, std::string const &Topic,
              SrcToDst Map, MessageWriter *Writer,
              Metrics::Registrar &RegisterMetric, time_point StartTime,
              duration StartTimeLeeway, time_point StopTime,
-             duration StopTimeLeeway)
+             duration StopTimeLeeway,
+             std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers)
```
SkyToGround, line 23:
```suggestion
             std::unique_ptr<Kafka::ConsumerFactoryInterface> CreateConsumers  = std::make_unique<Kafka::ConsumerFactory>())
```

PR #548, opened 2020-06-18
Make kafka config flag consistent with kafkacat

matthew-d-jones:
I suggest we change the flag for the Kafka config option from `-S` to `-X` to match kafkacat and kafkacow.


SkyToGround:
Looks like the system tests fails for whatever reason though.

PR #560, opened 2020-08-05
Re-factor HDFFile

SkyToGround:
### Description of work

* Split the (1100 line) *HDFFile.cpp*-file into three new files.
* Re-wrote the `HDFFile` class to inherit from a base class in order to better support unit testing.
* Modified unit tests and supporting code to accommodate the changes to the `HDFFile` class.
* Fixed bugs in writing of static data to the HDF file such that the code will now throw an error if the unit type is unknown instead of (as previously) fail silently.
* Moved code for checking the HDF5 version from being executed when starting a job to being executed when starting the application.


src/HDFOperations.h
```diff
@@ -0,0 +1,92 @@
...
+                              hdf5::node::Node const &Node,
+                              std::string const &Name, uint32_t StringSize,
+                              hdf5::datatype::CharacterEncoding Encoding,
+                              nlohmann::json const &Values,
+                              SharedLogger const &Logger);
+
+void addLinks(hdf5::node::Group const &Group, nlohmann::json const &Json,
+              SharedLogger Logger);
+
+}
```
SkyToGround, line 92:
```suggestion
}

```

src/HDFOperations.cpp
```diff
@@ -0,0 +1,859 @@
...
+    if (0 > H5Olink(TargetID, static_cast<hid_t>(Group), LinkName.c_str(),
+                    H5P_DEFAULT, H5P_DEFAULT)) {
+      Logger->warn("can not create link name: {}  in group: {}  to target: {}",
+                   LinkName, std::string(Group.link().path()), Target);
+      continue;
+    }
+  }
+}
+
+} // namespace HDFOperations
```
matthew-d-jones, line 859:
```suggestion
} // namespace HDFOperations

```

src/HDFFile.cpp
```diff
@@ -9,889 +9,66 @@
...
+                 nlohmann::json const &NexusStructure,
+                 std::vector<StreamHDFInfo> &StreamHDFInfo)
+    : H5FileName(FileName) {
 // Keep this.  Will be used later to test against different lib versions
 #if H5_VERSION_GE(1, 8, 0) && H5_VERSION_LE(1, 10, 99)
   unsigned int maj, min, rel;
   H5get_libversion(&maj, &min, &rel);
 #else
   static_assert(false, "Unexpected HDF version");
 #endif
```
matthew-d-jones, line 33: I suggest we take the opportunity to get rid of this unused code.
```suggestion
```

src/HDFOperations.cpp
```diff
@@ -0,0 +1,859 @@
...
+        {"float", [&]() { writeAttrNumeric<float>(Node, Name, Values); }},
+        {"double", [&]() { writeAttrNumeric<double>(Node, Name, Values); }},
+        {"string", [&]() {
+           writeAttrString(Node, Name, Values, StringSize, Encoding, Logger);
+         }}};
+    WriteAttrMap.at(DType)();
+  } catch (std::exception const &) {
+    auto ErrorStr =
+        fmt::format("Failed attribute write in {}/{} with data type {}",
+                    std::string(Node.link().path()), Name, DType);
```
matthew-d-jones, line 407: :+1: For `stringstream` -> `fmt`

src/HDFFile.cpp
SkyToGround: 👍

PR #564, opened 2020-09-21
DM-1651 Part 1. Remove redundant code, add helper function and move a function

SkyToGround:
### Description of work

* Removed duplicate code for getting the host name and PID.
* Added helper function for generating a random hex string (to be used in a later PR).
* Moved function for reading file into the only location where it is used.


matthew-d-jones (approved):
Should there be unit tests for `randomHexString`? Otherwise all looks fine to me.

SkyToGround:
@matthew-d-jones Sure, I will add a few unit tests later today.

PR #566, opened 2020-09-22
DM-1651: New command listener

SkyToGround:
### Description of work

* Moves and modifies the command listener.
* Suppresses unusedFunction-warnings to simplify use of small PRs.

### DIFF
--- a/src/CommandSystem/CommandListener.h
+++ b/src/CommandSystem/CommandListener.h
@@ -18,11 +18,14 @@ namespace Command {
 
 using FileWriter::Msg;
 
-/// Check for new commands on the topic, return them to the Master.
+/// \brief Check for new commands on the topic.
 class CommandListener {
 public:
   CommandListener(uri::URI CommandTopicUri, Kafka::BrokerSettings Settings);
 
+  /// \brief Poll the Kafka topic for a new command.
+  /// \note Will timeout on its fist call. Will continue to timeout as long as
+  /// it fails to retrieve metadata from the Kafka broker.
   std::pair<Kafka::PollStatus, Msg> pollForCommand();
 
 protected:
--- a/src/CommandSystem/CommandListener.h
+++ b/src/CommandSystem/CommandListener.h
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This code has been produced by the European Spallation Source
+// and its partner institutes under the BSD 2 Clause License.
+//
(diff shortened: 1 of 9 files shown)


src/CommandSystem/CommandListener.h
```diff
@@ -0,0 +1,39 @@
...
+
+using FileWriter::Msg;
+
+/// \brief Check for new commands on the topic.
+class CommandListener {
+public:
+  CommandListener(uri::URI CommandTopicUri, Kafka::BrokerSettings Settings);
+
+  /// \brief Poll the Kafka topic for a new command.
+  /// \note Will timeout on its fist call. Will continue to timeout as long as
```
SkyToGround, line 27:
```suggestion
  /// \note Will timeout on its first call. Will continue to timeout as long as
```

src/tests/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ set(UnitTests_SRC
         $<TARGET_OBJECTS:NeXusDatasetTests>
         StatusReporterTests.cpp
         helpers/StatusHelpers.cpp
-        HelperTests.cpp)
+        HelperTests.cpp CommandSystem/CommandListenerTests.cpp)
```
matthew-d-jones, line 44: Missing new line

matthew-d-jones (commented):
The approach to retry the metadata call on poll looks good to me :+1: 

I see an error message, and not the expected warning message, during the unit test
```
[15:25:02.594365] [error] [processID: 45805]: error at type: -1
```
Any idea what is going on there?

It would be nicer to pass the consumer to the command listener, so that you can pass a fake or mock and also test the happy path.

src/tests/CMakeLists.txt
SkyToGround:
```suggestion
        HelperTests.cpp CommandSystem/CommandListenerTests.cpp
        )
```

SkyToGround:
@matthew-d-jones We do get the expected warning and error messages based on the configured log level.
As for providing an external consumer. This is one of the cases where my guesstimate is that we need on the order of 100 lines of more code to enable that particular unit test. That together with the consequence and probability of failure in this class (low and low) makes me think that this is not worth it. Still, I can do it but I want to hear the opinion of someone else in that case as well.

matthew-d-jones (approved):
Sorry, I meant pass "consumer factory", I see that not constructing the consumer in the listener would require more work. If that doesn't change matters then I leave it to your discretion.

PR #567, opened 2020-09-22
Fix bug in Consumer::addPartitionAtOffset

SkyToGround:
### Description of work

* The `assign` call *replaces* topic-partitions instead of adding new ones. Thus when previously calling `addPartitionAtOffset`, you removed previous assignments. This PR fixes that.

### DIFF
--- a/src/Kafka/Consumer.cpp
+++ b/src/Kafka/Consumer.cpp
@@ -111,11 +111,19 @@ void Consumer::addPartitionAtOffset(std::string const &Topic, int PartitionId,
   Logger->info("Consumer::addPartitionAtOffset()  topic: {},  partitionId: {}, "
                "offset: {}",
                Topic, PartitionId, Offset);
-  auto TopicPartition = std::unique_ptr<RdKafka::TopicPartition>(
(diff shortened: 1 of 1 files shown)


src/Kafka/Consumer.cpp
```diff
@@ -111,11 +111,19 @@ void Consumer::addPartitionAtOffset(std::string const &Topic, int PartitionId,
...
+  auto ErrorCode = KafkaConsumer->assignment(Assignments);
+  if (ErrorCode != RdKafka::ERR_NO_ERROR) {
+    Logger->error("Could not assign to {}. Could not get current assignments.",
+                  Topic);
+    throw std::runtime_error(
+        fmt::format("Could not assign topic-partition of topic {}. Could not "
+                    "get current assignments. RdKafka error: \"{}\"",
+                    Topic, err2str(ErrorCode)));
+  }
+  Assignments.emplace_back(
```
matthew-d-jones, line 124: I think this should be `push_back`, otherwise everything looks good.

matthew-d-jones (approved):

PR #568, opened 2020-09-23
DM-1651 part 3, Add job listener

SkyToGround:
### Description of work

* A simple extension of `CommandListener` that adds some functionality required for listening to jobs.


matthew-d-jones (approved):
LGTM.
I also set `max.in.flight.requests.per.connection` to 1 in [pool-party](https://github.com/ess-dmsc/pool-party) but I _think_ it isn't required on the consumer side.

PR #573, opened 2020-10-15
Fix metrics bugs

SkyToGround:
### Description of work

* Fixes two separate bugs that prevented Graphite (Carbon) from parsing updates.
* Fixed one crashing bug from improperly de-registering Metrics counters.
* Added some missing code to the metrics registrar.


amues:
@SkyToGround, what were the Graphite bugs being fixed?

SkyToGround:
@amues Carbon was provided timestamps in ms instead of seconds. That was fixed. Furthermore, a null character was inserted at the end of each string which is incorrect as well.

amues (approved):

PR #574, opened 2020-10-20
Remove unused and misleading setting

matthew-d-jones:
`metadata.request.timeout.ms` is not actually used, so it was misleading that it is explicitly set, especially to very low value.
Metadata timeouts are specified in the metadata query function call.


SkyToGround (approved):

SkyToGround:
I am waiting for the build to complete without errors before I merge.

PR #577, opened 2020-11-16
Better hdf attributes

SkyToGround:
### Description of work

- All writer modules now set a *NX_class* value.
- The *NX_class* value configured for a writer module will always override that of the HDF group *NX_class* value. If configured, a writer module attribute (e.g. *NX_class*) will override both.
- The Kafka topic and flatbuffer source names will now automatically be written as (HDF) attributes when instantiating a new writer module.
- Minor fix to CMake code for working around issues caused by differences in how file-name leading lower and upper case letters are handled on different systems.


zjttoefs (approved):
Looking at the code on Github I see no issues. I assume the tests complete successfully. So we're good to go.

PR #578, opened 2020-12-01
ECDC-1719 Fix writer module configurations

SkyToGround:
### Issue

* ECDC-1719 (kind of)

### Description of work

I pretty happy about this one. Disregarding the added unit tests and documentation, it is roughly the same number of lines of code as before but with a lot more functionality. In the process I have been fixing SEVERAL long standing grievances I have had with the file-writer. The non-detailed version of the changes are:

- The writer module configuration names/keys have been unified.
- Better documentation of writer module configuration options.
- The application will now print an error message if there is a configuration that is not used (due to e.g. a typo).
- The error reporting and handling of writer module configurations have overall been greatly improved.

Some additional notes:
- Have made some minor fixes to some of the writer modules including removal of useless functionality.
- Due to the previous item, I have removed a few unit tests.
- The hs00 module is **a mess**. It should (probably) either be re-written or dropped entirely as it seems that no-one is using it anyway. I started improving it but then thought better.


zjttoefs:
Looks like a good clean-up and I like how things have been unified.
Before I look closer into this: Why is .process_config() needed? 
It is apparently always called after .parse_config(<>). Why make that a two-step process?

SkyToGround:
`parse_config()` is for doing just that and is now not to be touched by the developer of a writer module. However, in some cases additional processing of some configuration field is required (e.g. converting a string to a data type) in which case this can be done in `process_config()`. This could of-course be baked into `init_hdf()` but that in that case the code  would in some circumstances have to be called from `reopen()` as well. Is is better in my opinion to separate responsibility and let it be its own function.

src/WriterModuleConfig/FieldHandler.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <nlohmann/json.hpp>
+#include <set>
+
+namespace WriterModuleConfig {
+
+using nlohmann::json;
+
+void FieldHandler::registerField(FieldBase *Ptr) {
+  auto Keys = Ptr->getKeys();
+  for (auto Key : Keys) {
```
matthew-d-jones, line 22:
```suggestion
  for (auto const &Key : Keys) {
```

src/WriterModuleConfig/FieldHandler.cpp
```diff
@@ -0,0 +1,72 @@
...
+      LOG_WARN(
+          "Replacing the config field (key) \"{}\". Note: this is programming "
+          "error (i.e. a bug) that should be fixed post-haste.",
+          Key);
+    }
+    FieldMap[Key] = Ptr;
+  }
+}
+
+void FieldHandler::processConfigData(std::string ConfigJsonStr) {
```
matthew-d-jones, line 33:
```suggestion
void FieldHandler::processConfigData(std::string const &ConfigJsonStr) {
```

src/WriterModuleConfig/Field.h
```diff
@@ -0,0 +1,106 @@
...
+class FieldBase {
+public:
+  FieldBase(WriterModule::Base *Ptr, std::vector<std::string> const &Keys);
+  FieldBase(WriterModule::Base *Ptr, std::string const &Key)
+      : FieldBase(Ptr, std::vector<std::string>{Key}) {}
+  virtual ~FieldBase() {}
+  virtual void setValue(std::string const &NewValue) = 0;
+  bool hasDefaultValue() const { return GotDefault; }
+  auto getKeys() const { return FieldKeys; }
+  bool isRequried() const { return FieldRequired; }
```
matthew-d-jones, line 33:
```suggestion
  [[nodiscard]] bool hasDefaultValue() const { return GotDefault; }
  [[nodiscard]] auto getKeys() const { return FieldKeys; }
  [[nodiscard]] bool isRequried() const { return FieldRequired; }
```

src/WriterModuleConfig/Field.h
```diff
@@ -0,0 +1,106 @@
...
+        FieldType DefaultValue)
+      : FieldBase(WriterPtr, Keys), FieldValue(DefaultValue) {}
+
+  void setValue(std::string const &ValueString) override {
+    setValueImpl<FieldType>(ValueString);
+  }
+
+  FieldType getValue() { return FieldValue; };
+
+  operator FieldType() const { return FieldValue; }
```
matthew-d-jones, line 59:
```suggestion
  explicit operator FieldType() const { return FieldValue; }
```

matthew-d-jones (changes requested):
Nice job with the cleanup and docs.
A few minor suggestions below.

src/CMakeLists.txt
```diff
@@ -105,7 +105,7 @@ set(kafka_to_nexus_SRC
         CommandSystem/CommandListener.cpp
         CommandSystem/JobListener.cpp
         TimeUtility.cpp
-        )
+        WriterModuleConfig/Field.cpp WriterModuleConfig/Field.h WriterModuleBase.cpp)
```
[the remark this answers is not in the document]
SkyToGround: This or similar alternatives have been suggested multiple times in the past but no-one has thought it to be enough of a problem to implement it. I would not mind if you did though. :-)

src/WriterModule/f142/f142_Writer.cpp
```diff
@@ -70,16 +51,15 @@ void initValueDataset(hdf5::node::Group &Parent, Type ElementType,
...
-    Parent["value"].attributes.create_from<std::string>("units", *ValueUnits);
-  }
 }
 
 /// Parse the configuration for this stream.
-void f142_Writer::parse_config(std::string const &ConfigurationStream) {
-  auto ConfigurationStreamJson = json::parse(ConfigurationStream);
-
+void f142_Writer::process_config() {
+  auto toLower = [](auto InString) {
```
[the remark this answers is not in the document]
SkyToGround:
```suggestion
  auto ToLower = [](auto InString) {
```

src/WriterModuleConfig/FieldHandler.h
```diff
@@ -0,0 +1,29 @@
...
+public:
+  FieldHandler() = default;
+  void registerField(FieldBase *Ptr);
+  void processConfigData(std::string ConfigJsonStr);
+
+private:
+  std::map<std::string, FieldBase *> FieldMap;
+};
+
+} // namespace WriterModuleConfig
```
SkyToGround, line 29:
```suggestion
} // namespace WriterModuleConfig

```

src/WriterModuleConfig/Field.h
```diff
@@ -0,0 +1,106 @@
...
+
+template <class FieldType> class RequiredField : public Field<FieldType> {
+public:
+  RequiredField(WriterModule::Base *WriterPtr, std::string const &Key)
+      : Field<FieldType>(WriterPtr, Key, FieldType()) {
+    FieldBase::makeRequired();
+  }
+};
+
+} // namespace WriterModuleConfig
```
SkyToGround:
```suggestion
} // namespace WriterModuleConfig

```

src/WriterModuleConfig/Field.h
SkyToGround: In this particular case: no. Fixing all the implicit conversions (that has been there from the beginning) is a part of another PR, not this one.

SkyToGround:
@zjttoefs Based on our discussion, I have cleaned up the use of `process_config()` (now `config_post_processing()`) from most places. There are ways of getting rid of it completely. One of them is to have the JSON library do the conversion/processing directly. This turned out to be surprisingly tricky though and when I attempted it, half an hour was not enough to get it to work properly. The other alternative is to have `class Field<>` call a conversion function. Both of these alternatives are cleaner than the current alternative but I would argue that it makes code discoverability (especially for junior developers) worse. But then again, who cares about those guys. :-)

src/WriterModuleConfig/Field.h
matthew-d-jones: ok

src/CMakeLists.txt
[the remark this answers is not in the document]
matthew-d-jones: Created ticket ECDC-2243, a nice easy ticket for me to do on a Friday afternoon at some point perhaps.

matthew-d-jones (approved):
I'm happy at this point, haven't merged to give @zjttoefs opportunity to take another look if he wishes.

zjttoefs:
I like the looks of it now! Meets my needs for being discoverable and I especially like that you don't need to call the second method externally any more. Nice.

Since we discussed it, why don't we do the three classes for _default_ chunk sizes?

ev42 + hs00 + NDAr + ADAr "x-large"
That would mean to increase ev42 to 1M. hs00 looks okay. I did not check the others.

tdctc + senv "large"
4k seems fine

"everything else" (ns10, ns11, f142, ...)
1k

I would also be happy to merge now, if that discussion delays things.

SkyToGround:
Default chunk size updated. I will merge as soon as the build system works again and the build passes.

zjttoefs (changes requested):
Chunk sizes in code and documentation are out of sync.
Will merge when fixed (and build works).

zjttoefs (approved):

PR #579, opened 2020-12-17
Minor changes to the writer config field functionality

SkyToGround:
### Description of work

* Changed the name(space) off the config field.
* Un-coupled the config field code from the writer module.
* Created `KeyString` type to avoid bugs caused by implicit use of `std::string`-constructor.
* Improved logging by adding code for the _fmt_ library to print _JSON_-objects.


matthew-d-jones:
Moving `FieldHandler` to the `JsonConfig` namespace to make use of it elsewhere makes sense, but I don't understand the rest of the changes. It seems to achieve the opposite of what you put in the description. Coupling with writer modules has been increased by switching to inheritance from composition; changes to the interface of `FieldHandler` can now impact everything that uses `WriterModule::Base`.

matthew-d-jones (commented):
Moving `FieldHandler` to the `JsonConfig` namespace to make use of it elsewhere makes sense, but I don't understand the rest of the changes. It seems to achieve the opposite of what you put in the description. Coupling with writer modules has been increased by switching to inheritance from composition; changes to the interface of `FieldHandler` can now impact everything that uses `WriterModule::Base`.

SkyToGround:
If you want to, you could perhaps argue that `WriterModule::Base` is more tightly coupled to `FieldHandler` though I do not really agree with that notion. It is a clear fact though that `FieldHandler` is now less coupled (totally uncoupled in-fact) with `WriterModule::Base` and that is the intent with this change.

I noticed though that I had not changed the log messages and I will do so now.

matthew-d-jones (commented):
I don't feel I can approve this PR. Switching from composition to inheritance for `WriterModule::Base` and `FieldHandler` looks like a regression to me.
We could invite comment from others in case they disagree with me or think it is not worth rejecting these changes over. @mattclarke @zjttoefs

SkyToGround:
@matthew-d-jones As long as you agree that the PR that you are rejecting does indeed reduce coupling I am ok with that. There is at least one somewhat more complicated solution (as well as an ugly one) that can (probably) solve the problem as well.

zjttoefs:
The PR title mentions a change to config field functionality. What functionality is changing?
 
Looks like the very first commit (65ac4f648a44b9019715ebfc03199f04766cdbec) does not do anything like that, but moves `FieldHandler` to the `JsonConfig` namespace. That is the part that everyone is happy with, I think. Despite the misleading title.

According to the text of the PR the aim is to decouple the class hierarchy. Something that @matthew-d-jones does not appear to agree that it is happening. I don't see it either. In https://github.com/ess-dmsc/kafka-to-nexus/pull/579#issuecomment-772132040 you (@SkyToGround) mention a potential other solution. To what problem?

SkyToGround:
@matthew-d-jones was confused (I assume) over the difference betwen coupling and inheritance. However, it seems to me that he realised what was going on and pointed out that he did not like the use of inheritance in this PR. This particular comment has now been adressed by replacing inheritance with a bunch of template-code.

SkyToGround:
Oh, and seeing as this PR was created almost two months ago, several changes have been piled on to it since the title was written. I will add a list to the description.

zjttoefs:
Looks like we are agreeing that there is no functionality change, then. Other than maybe for logging.
I'll let @matthew-d-jones comment on the inheritance for the fields. At a glance that looks like his concerns may be addressed.
Deriving KeyString from std::string appears to come with warnings by some C++ experts. I can't say it those potential problems are worse than the potential problems you are trying to fix with that scheme. Do you have a comment?

SkyToGround:
You mean comment on `KeyString`? Inheriting from a base class that does not have a virtual destructor can indeed be dangerous. Due to me not adding any functionality to `KeyString`, that does not apply in this particular case.

matthew-d-jones:
Firstly, WriterModule+Field:
I wonder if the source of confusion here is that the relationship previously was not simply composition; there are various pointers being passed around... regardless it doesn't look like it is entangling them further now, so let's go with this implementation.

Secondly, KeyString:
Where did you get into trouble with std::string? Can we avoid that problem rather than introducing KeyString?
My motivation for seeking to avoid KeyString is that having something slightly different from std::string in the codebase is adding complexity/confusion for other developers, and that inheriting from a std class is setting a bad precedent (even though, as you've pointed out, it won't cause a problem exactly as it is implemented now).

SkyToGround:
@matthew-d-jones Ah, you did not previously mention that you did not like `KeyString` as far as I can tell. Although I have mentioned in the code comments why I introduced `KeyString` I can clarify it a bit here. The problem is that this code will compile without warning: `std::string SomeString{"alt1", "alt2"}`. Note also the declaration for the two constructors in `FieldBase` and I am sure that you can see why this is an issue.

I can of course re-write `KeyString` to use composition as well if you really feel that it is required but I think it is a bit of waste of time and code.

matthew-d-jones:
@SkyToGround  Are you saying that something like this compiles for you?
```
JsonConfig::Field<int> UnderTest(FieldHandler.get(), {"h1", "h2"}, 33);
```
gcc gives me a nice clear error that it is ambiguous which constructor I want and that I therefore need to do
```
JsonConfig::Field<int> UnderTest(FieldHandler.get(), std::vector<std::string>{"h1", "h2"}, 33);
```

matthew-d-jones:
I've just checked on godbolt, I get a nice clear ambiguous constructor error on clang too
https://godbolt.org/z/3b1r9x

SkyToGround:
I had to go back and find some old commits to remember exactly what happened. Take a look at this line:
```
JsonConfig::Field<int> UnderTest(FieldHandler.get(), {{"h1", "h2"}}, 33);
```
This has happened twice from what I remember so `KeyString` in one form or another stays.

matthew-d-jones (approved):
When you've had a problem with the normal, expected behaviour of the language and standard library, which happens to us all, I don't think the right response is to introduce code to make it non-standard and create more room for confusion and other problems down the line.
I don't think we'll reach agreement on this, so I will mark the PR as accepted because it is otherwise functioning fine and has already taken too much of our time.

Tagging @zjttoefs so that he is aware of this conclusion.

SkyToGround:
Thanks for the approval. C++ is a big and complicated language and there plenty of areas where it is simply poorly designed (or in this case, inherited from C). Automatic implicit type casting being one of them. I do not consider "the language is bad, just get good" to be a good argument.

PR #583, opened 2021-01-27
Fix timeout/eop issue

SkyToGround:
### Issue

* ECDC-2170

### Description of work

This PR *might* fix the issue of discerning between EOP and timeout. However, it is somewhat hard to test due to the current number of outstanding PRs so I have not done so and thus can not say if it is an improvement.

### Reminder


zjttoefs:
Looks like this failed the automated test. Is this ready for review?

SkyToGround:
Yes it is ready for review. The failure is (as far as I can tell) a bug in cpp-check. I have disabled that failing test in master and I will apply that fix here as well.

SkyToGround:
Turns out that there is a system-test issue as well. It is not worth fixing that now as that code will soon (maybe/hopefully) will be updated with the new system tests. I still would not mind if you want to review the code but the fix of the system-test issue will have to wait for another few months it appears.

matthew-d-jones (commented):
Implementation looks fine to me and unit tests make sense.

Can you remind me what the behaviour is before the change, did you put another mechanism in place to avoid it waiting indefinitely in the case that no message is ever sent after the stop time? Or is that a problem that could occur and system tests were adjusted to avoid it?

SkyToGround:
The behaviour was that without knowledge of EOP, timeouts had to be treated as EOP and thus connection issues (which results in timeouts) could result in writing stopping before we reached the stop time on a particular topic.

matthew-d-jones:
@SkyToGround oh yes, thanks. I think this change is pretty safe then, I'll mark as accepted.
Do you have any idea what is causing the system test problem, we don't seem to be getting this issue on the Forwarder tests?

matthew-d-jones (approved):

SkyToGround:
No, but I am starting to suspect that it is an issue with docker.

matthew-d-jones:
@SkyToGround
Sounds very plausible, perhaps on one particular system test node
I've found this Forwarder PR hit the same issue:
https://jenkins.esss.dk/dm/blue/organizations/jenkins/ess-dmsc%2Fforwarder/detail/PR-65/6/pipeline

zjttoefs (approved):

zjttoefs:
Looks good. I would have merged, but the merge build fails and I don't want to deal with that fallout...

SkyToGround:
Me neither, I will let this PR wait for now.

PR #588, opened 2021-03-23
Development

SkyToGround:
### Issue

*Specify the associated JIRA ticket, e.g. DM-1234.*

### Description of work

*Add a description of the work done. The aim is provide information to help the reviewer review the PR efficiently.
See [CONTRIBUTING.MD](CONTRIBUTING.MD) for more information on code standards etc.*

### Reminder


src/Stream/Topic.h
```diff
@@ -98,7 +98,9 @@ class Topic {
   void checkIfDone();
   virtual void checkIfDoneTask();
 
-  bool shouldGiveUp(std::string const &GiveUpMessage);
+  virtual bool shouldGiveUp(std::string const &GiveUpMessage);
```
zjttoefs, line 101: That looks like an odd method signature. The two times it is used it uses the same message string. It also modifies the state of the object (sets the error state to true). Something that is unexpected from the boolean method. 
I'd prefer if the method just constrained itself to the time comparison.

Also not sure why there is a need to override the behaviour.

SkyToGround: We could/can discuss your comments in detail when next time we meet if you want. Your line of questioning is valid though it leads to the eternal problem in actual real world unit testing (as opposed to the pie in the sky unit testing that some people fantasise about): Do we allow (some of) the unit tests to be aware of the implementation or do we ignore testing some functionality?

In short, is there some better implementation for doing what we want here? Probably. Is it worth it to re-factor/re-implement this class to get to it? Possibly. But I doubt it until I or some one else has an idea of what the better implementation might be.

SkyToGround: Oh, I missed the identical strings thing. It is fixed now.

zjttoefs: So I guess that answers the question that I added last to my comment: The overriding is there for the unit test's benefit? I might not be too fond of that, but we could talk about that.
What I like event less after a night of marinating over it is the method signature:
* The name  `shouldGiveUp` is not clear if that retrieves information or sets a state. I'd go with `checkMetadataTimeout`
* The boolean nature makes it look like you query it and can initiate action based on the return value, but in fact it takes most of the required action (except for the top level `return` it can't do).

I'd rather extract the latter like so:
```
void Topic::setErrorState(const std::string &GiveUpMessage) {
    HasError = true;
    std::lock_guard Lock(ErrorMsgMutex);
    ErrorMessage = GiveUpMessage;
    LOG_ERROR(ErrorMessage);
    return;
}
```

SkyToGround: Right, I have made changes to the code based on your suggestion. I veto renaming `shouldGiveUp` to `checkMetadataTimeout` though as "metadata timeout" is already used in that class for setting the metadata timeout of blocking metadata calls.

PR #597, opened 2021-07-26
ECDC-2618 Updating error messages in FileWriterTask class

kmurica:
### Issue

ECDC-2618

### Description of work

Minor updates to handling of error messages in FileWriterTask for the two cases:

1. Nexus file already exists.
2. Directory provided in .ini file does not exist.

### DIFF
--- a/src/FileWriterTask.cpp
+++ b/src/FileWriterTask.cpp
@@ -35,7 +35,7 @@ std::vector<Source> &FileWriterTask::sources() { return SourceToModuleMap; }
 
 void FileWriterTask::setFilename(std::string const &Prefix,
                                  std::string const &Name) {
-  FilePath = Prefix;
+  FileDirectory = Prefix;
   if (Prefix.empty()) {
     Filename = Name;
   } else {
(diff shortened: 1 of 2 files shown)


src/FileWriterTask.cpp
```diff
@@ -49,13 +50,23 @@ void FileWriterTask::addSource(Source &&Source) {
 void FileWriterTask::InitialiseHdf(std::string const &NexusStructure,
                                    std::vector<StreamHDFInfo> &HdfInfo) {
   auto NexusStructureJson = hdf_parse(NexusStructure, Logger);
+  std::string ErrorString;
+  if (std::filesystem::exists(Filename)){
+    ErrorString = fmt::format("Failed to initialize HDF file \"{}\". Error was: \"{}\"", Filename,
+                              "A file with that filename already exists in that directory. Delete the existing file or provide another filename.");
+    std::throw_with_nested(std::runtime_error(ErrorString));
+  }
+  else if(not FileDirectory.empty() and not std::filesystem::exists(FileDirectory)){
```
umithardal, line 59: Why do we need the dir to be empty? Or this checks something else?

umithardal:
other than the above question, LGTM.

src/FileWriterTask.cpp
kmurica: it should not be empty. Apparently you are not obliged to provide a directory to the file writer (I suppose it just writes it in the bin directory directly then and that will exist).

umithardal (approved):

PR #598, opened 2021-07-27
ECDC-2611: File writer system tests to use confluent_kafka

ebadkamil:
### Description of work

* Start using confluent_kafka library in system tests.
* wait_until_kafka_ready method in conftest, now uses producer to send message to kafka and waits until the message delivery was acknowledged (as it is done in forwarder library's system tests) and then only declares that kafka is ready. Old way of successfully creating consumer and declaring kafka is ready **seems to create spurious and random errors while running system-tests locally**.  

NOTE: This PR does not fix assertion error in `test_filewriter_static_data.py` . That will be a different ticket.

### DIFF
--- a/system-tests/conftest.py
+++ b/system-tests/conftest.py
@@ -1,14 +1,16 @@
 import os.path
+import signal
+import time
+import warnings
+from subprocess import Popen
+
+import docker
 import pytest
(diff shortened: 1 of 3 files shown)


kmurica (approved):
Looks good to me. Works much better on my local machine now! Good job.

PR #599, opened 2021-08-03
Sprint2107

kmurica:
### Issue

Sprint 2107 work

### Description of work

Changes in the FW in the scope of the sprint 2107

Jonas (SkyToGround)


SkyToGround (commented):
The `Name` can also contain parts of a path and thus you get the full path from taking both the `Prefix` and the `Name`. Thus, I suggest that instead of going this route, you re-implement the code using `std::filesystem::path` and that using this object you do the following:
1. Check that it is a valid path.
2. Check that it has a valid file-name.
3. Check that the parent path (the directory) exists.
4. Check if the file already exists.

SkyToGround (approved):
Looks good.

PR #603, opened 2021-10-06
Re-activated documentation failure code.

SkyToGround:
### Issue

- ECDC-1654

### Description of work

...


kmurica (approved):
Good idea. I think this is the only reasonable way to deal with this ticket.

PR #615, opened 2021-11-24
removing non-job pool mode

kmurica:
### Issue

ECDC-2619

### Description of work

Removing the non-job pool way to run file writer as we want to move away from having two separate ways to run the filewriter.
It should be possible to provide the same topic name for both the job pool and and command status
command-status-uri=localhost:9092/Loki_writerCommand
job-pool-uri=localhost:9092/Loki_writerCommand

System tests are updated to only use the job pool mode and nothing else.
There are also updates to the file-writer-control library found in this PR:
https://github.com/ess-dmsc/file-writer-control/pull/39

### DIFF
--- a/src/MainOpt.h
+++ b/src/MainOpt.h
@@ -57,7 +57,7 @@ struct MainOpt {
   uri::URI CommandBrokerURI{"localhost:9092/kafka-to-nexus.command"};
 
   /// Kafka broker and topic where file writer jobs are published.
-  uri::URI JobPoolURI{""};
+  uri::URI JobPoolURI{"localhost:9092/kafka-to-nexus.jobpool"};
 
   /// \brief Path for HDF output.
   ///
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ Part of the ESS data streaming pipeline.
   --version                   Print application version and exit
   --command-status-uri URI REQUIRED
                               <host[:port][/topic]> Kafka broker/topic to listen for commands and to push status updates to.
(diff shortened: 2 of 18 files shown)


michele-brambilla:
Can we please discuss this change before proceed to merge?

kmurica:
> Can we please discuss this change before proceed to merge?

Sure, there are also the ANSTO facility that needs to be inquired.

src/MainOpt.h
```diff
@@ -57,7 +57,7 @@ struct MainOpt {
   uri::URI CommandBrokerURI{"localhost:9092/kafka-to-nexus.command"};
 
   /// Kafka broker and topic where file writer jobs are published.
-  uri::URI JobPoolURI{""};
+  uri::URI JobPoolURI{"localhost:9092/kafka-to-nexus.jobpool"};
```
SkyToGround, line 60: I would really prefer that you either remove the default address here and/or make the corresponding command line argument required (it has the same effect I believe). You can make the CL argument required on line 137 of "CLIOptions.cpp".

SkyToGround: My previous statement was wrong. You have to make the command "required".

SkyToGround (changes requested):
One minor CLI change is also needed.

SkyToGround:
One more thing that I forgot: Make the appropriate changes to the documentation and to the "changes.md"-file (maybe under a heading called "Next version" or something similar).

geishm-ansto:
@kmurica We have used the writer in the job pool mode at Ansto and am comfortable with the change.

PR #616, opened 2021-12-02
No longer stops a job if the start time is too old

mattclarke:
Instead just warns


zjttoefs:
Makes total sense. 

I am curious to find out what happens if the request goes back so far in time the records in Kafka become patchy or non-existent. But that's a different issue.

PR #652, opened 2022-10-10
ECDC-2999: Adding full path to file writing done message

kmurica:
### Issue

Closes https://jira.esss.lu.se/browse/ECDC-2999

### Description of work

Adding full file path to finished file writing message.

### Reminder

### DIFF
--- a/src/Master.cpp
+++ b/src/Master.cpp
@@ -148,14 +148,11 @@ std::string Master::getCurrentFileName() const {
   return CurrentStatus.Filename;
 }
 
-std::string Master::getCurrentFilePath() const {
+std::filesystem::path Master::getCurrentFilePath() const {
   std::lock_guard LockGuard(StatusMutex);
-  std::string fullFilePath = MainConfig.getHDFOutputPrefix();
-  if (fullFilePath.empty()) {
-    fullFilePath = CurrentStatus.Filename;
-  } else {
-    fullFilePath = fullFilePath + "/" + CurrentStatus.Filename;
-  }
+  std::filesystem::path fullFilePath = MainConfig.getHDFOutputPrefix();
+  fullFilePath.append(CurrentStatus.Filename);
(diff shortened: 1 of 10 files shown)


kmurica:
Follow-up ticket will be created to address proposal/instrument based sub-directories.

src/Master.cpp
```diff
@@ -148,6 +148,17 @@ std::string Master::getCurrentFileName() const {
   return CurrentStatus.Filename;
 }
 
+std::string Master::getCurrentFilePath() const {
+  std::lock_guard LockGuard(StatusMutex);
+  std::string fullFilePath = MainConfig.getHDFOutputPrefix();
+  if (fullFilePath.empty()) {
+    fullFilePath = CurrentStatus.Filename;
+  } else {
+    fullFilePath = fullFilePath + "/" + CurrentStatus.Filename;
```
carez, line 157: For building the path in a more portable manner, consider using std::filesystem::path and the `/` operator: https://en.cppreference.com/w/cpp/filesystem/path/append

carez (commented):
The code looks good to me apart from the small comment below.

However, I wonder if this is an "anti-feature". IMHO the full path where files are stored is an implementation detail of the storage system.
The relative path is the unique identifier of a file, no matter where the file system decides to store it.

Components that make direct use of the filesystem should receive the HDFOutputPrefix as a configuration parameter (like the filewriter already does). But any other component should deal with the relative path as the unique identifier.

If I understand correctly, we want to provide the full path to SciCat_ingestor, but I don't see why the ingestor should know this. The ingestor will need a companion service that will be able to take one relative path and serve it via HTTP or some other protocol. The companion service will need access to the filesystem, but SciCat_ingestor not.

kmurica:
> 

Further discussion can be read in ticket https://jira.esss.lu.se/browse/ECDC-2999

carez (approved):
LGTM

PR #653, opened 2022-10-11
ECDC-3107: honour broker settings

carez:
### Issue

https://jira.esss.lu.se/browse/ECDC-3107

### Description of work

kafka-to-nexus performs a number of Kafka operations assuming that a broker address is enough to establish a connection. This is however not true for SSL or SASL connections, where librdkafka parameters read from `-X` flags must be used when creating the Kafka Handle.

This PR passes around the BrokerSettings (the ones created by parsed from command line parameters) to all parts of the code that create Kafka connections.

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -9,6 +9,7 @@
 - The following dependencies have been updated:
   - graylog-logger ([#650](https://github.com/ess-dmsc/kafka-to-nexus/pull/650))
 - Enabled SSL and SASL in librdkafka to support Kafka authentication.
+- Fix to make all Kafka connections honour the provided librdkafka parameters.
 
 ## Version 5.1.0: Attributes and dependencies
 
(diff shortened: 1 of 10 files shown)


carez:
@kmurica Here some examples where the code was receiving KafkaSetings but then overriding the Broker Address with values from `MainConfig` or elsewhere:

- https://github.com/ess-dmsc/kafka-to-nexus/blob/98de7fad1549b8e5738b7aa633428bf3786aa4d0/src/Kafka/ConsumerFactory.cpp#L31
- https://github.com/ess-dmsc/kafka-to-nexus/blob/98de7fad1549b8e5738b7aa633428bf3786aa4d0/src/kafka-to-nexus.cpp#L72
- https://github.com/ess-dmsc/kafka-to-nexus/blob/98de7fad1549b8e5738b7aa633428bf3786aa4d0/src/CommandSystem/CommandListener.cpp#L24
- This one looks more a specific feature: https://github.com/ess-dmsc/kafka-to-nexus/blob/98de7fad1549b8e5738b7aa633428bf3786aa4d0/src/CommandSystem/FeedbackProducer.cpp#L19

kmurica (approved):
After looking at it for a while, I do not think we can simplify it much more than this at the moment.
Looks good to me.

PR #657, opened 2022-10-28
ECDC-2967: Continue listening to inactive topic

kmurica:
### Issue

https://jira.esss.lu.se/browse/ECDC-2967

### Description of work

The file writer does not unsubscribe from the Kafka topic anymore when messages are not receieved
after a certain configurable time (default 30s). Instead a warning message is logged (only once per occurrence).

Jenny Walker


JennyLouise (approved):
All looks good to me

PR #659, opened 2022-11-02
ECDC-3142 Ignore broker addresses sent in JobStart message

danesss:
### Issue

https://jira.esss.lu.se/browse/ECDC-3142

### Description of work

We want to decouple the Kafka addresses seen by Nicos and the Filewriter. In this commit we stop honouring the address sent by Nicos and use one of the addresses already known by the Filewriter.

Note that this is a temporary solution to mitigate the coupling with Nicos.
We want to refactor the Kafka configuration in the filewriter (see ECDC-3118), possibly resulting in a single "kafka-config" for all librdkafka settings.
The code in this commit should use such new configuration facility to set the broker address instead of using the address from JobPoolURI, which is quite arbitrary.

### Reminder

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -2,6 +2,8 @@
 
 ## Next version
 
+- Breaking: Ignore Kafka IP addresses sent in `StartJob` messages, experiment
+  data is now fetched from the broker configured in `job-pool-uri`.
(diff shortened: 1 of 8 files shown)


kmurica (approved):
Looks good to me!

PR #667, opened 2022-12-12
minor fix to se00 module

kmurica:


hurvan (approved):
That looks good to me!

PR #670, opened 2023-01-13
Redact Kafka SASL credentials in logs

amues:
### Issue

ECDC-3152.

### Description of work

Redact Kafka SASL password in logs. Also add CMake configuration for macOS build workaround due to warning/error generated by asio code.

### Nominate for Group Code Review

- [ ] Nominate for code review 

### Reminder

*Changes should be documented in `changes.md`*

PR #672, opened 2023-01-27
Fix redaction of sensitive settings inadvertently modifies the setting

danesss:
### Issue

ECDC-3152

### Description of work

Fixes bug introduced in PR https://github.com/ess-dmsc/kafka-to-nexus/pull/670.
Sensitive configuration settings were not only redacted in logs, but the redacted value was also used to configure the application.


src/Kafka/ConfigureKafka.h
```diff
@@ -9,6 +9,8 @@
 
 #pragma once
 
+#include <regex>
```
mattclarke, line 12: Do you really need this in the header?

PR #686, opened 2023-04-03
Throttle Kafka consumption to the file writing speed

danesss:
### Issue

ECDC-3253: We currently read from Kakfa at max speed and buffer into memory, which can cause out-of-memory issues if Kafka is faster than the GPFS filesystem. 

### Description of work

We adapt the consumption rate to the file writing rate:

- We add a new configuration setting for the size of write queue `--max-queued-writes`
- We add a method in MessageWriter to be able to check current queue size
- We extend StreamController to
  - Perform a periodic check of the write queue size
  - Pause all consumers for 200ms if queue is larger than configured
- `Topic` and `Partition` classes have new methods to support `pause` and `resume` of the consumers.
- Unit tests are added to verify pause/resume behaviour at the `Partition` level.
  - To test the "queue_size -> pause" logic further changes are needed so they will be explored in separate PR: https://github.com/ess-dmsc/kafka-to-nexus/pull/689

### Reminder

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,13 @@ Part of the ESS data streaming pipeline.
   --data-flush-interval       (Max) amount of time between flushing of data to
                               file, in seconds.  Ex. "10s". Accepts "h", "m",
                               "s" and "ms".
+  --max-queued-writes         Maximum number of messages buffered for writing.
+                              Directly affects the memory usage of the
+                              application. The maximum is not enforced, only
+                              used as guideline to throttle Kafka consumption.
+                              Note that total memory usage will also depend on
+                              the size of the actual messages consumed from
+                              Kafka.
   -X,--kafka-config KEY VALUE ...
                               LibRDKafka options
   -c,--config-file            Read configuration from an ini file
--- a/changes.md
+++ b/changes.md
@@ -2,6 +2,9 @@
 
 ## Next version
 
+- Feat: Cap the size of the output buffer (`max-queued-writes` setting) to
+  avoid out-of-memory issues. Kafka consumers are paused until the buffer
+  usage falls below a fixed threshold.
 - Adding (tcp based) service api to query a filewriter for its status
 - Fix: Cannot import extra modules due to mismatch in ep00, ep01 and al00 registered names
 


src/Stream/Partition.h
```diff
@@ -52,6 +52,19 @@ class Partition {
...
+  /// \brief Pause the consumer thread.
+  ///
+  /// Non blocking. Will stop the queuing of poll commands to the consumer
+  /// thread command queue. There are no guarantees for when the consumer is
+  /// actually paused.
+  void pause();
+
+  /// \brief Resume the consumer thread.
+  ///
+  /// Non blocking. Will resume the queuing of poll commands to the consumer
```
mattclarke, line 64:
```suggestion
  /// Non-blocking. Will resume the queuing of poll commands to the consumer
```

src/Stream/Partition.h
```diff
@@ -52,6 +52,19 @@ class Partition {
   /// \note This function exist in order to make unit testing possible.
   void start();
 
+  /// \brief Pause the consumer thread.
+  ///
+  /// Non blocking. Will stop the queuing of poll commands to the consumer
```
mattclarke, line 57:
```suggestion
  /// Non-blocking. Will stop the queuing of poll commands to the consumer
```

src/Stream/Topic.h
```diff
@@ -36,6 +36,19 @@ class Topic {
   /// \note This function exist in order to make unit testing possible.
   void start();
 
+  /// \brief Pause the consumer threads.
+  ///
+  /// Non blocking. Will tell the consumer threads to pause as soon as possible.
```
mattclarke, line 41:
```suggestion
  /// Non-blocking. Will tell the consumer threads to pause as soon as possible.
```

src/Stream/Topic.h
```diff
@@ -36,6 +36,19 @@ class Topic {
...
 
+  /// \brief Pause the consumer threads.
+  ///
+  /// Non-blocking. Will tell the consumer threads to pause as soon as possible.
+  /// There are no guarantees for when the consumers are actually paused.
+  void pause();
+
+  /// \brief Resume the consumer threads.
+  ///
+  /// Non blocking. Will tell the consumer threads to resume as soon as
```
mattclarke, line 47:
```suggestion
  /// Non-blocking. Will tell the consumer threads to resume as soon as
```

src/StreamController.cpp
```diff
@@ -157,4 +174,22 @@ void StreamController::checkIfStreamsAreDone() {
   }
 }
 
+void StreamController::throttleIfWriteQueueIsFull() {
+  auto QueuedWrites = WriterThread.nrOfWritesQueued();
+  if (QueuedWrites > StreamerOptions.MaxQueuedWrites &&
+      !StreamersPaused.load()) {
+    LOG_DEBUG("Maximum queued writes reached (count={}). Pausing consumers...",
```
mattclarke, line 181:
```suggestion
    LOG_DEBUG("Maximum queued writes exceeded (count={}). Pausing consumers...",
```

src/StreamController.h
```diff
@@ -92,15 +105,26 @@ class StreamController : public IStreamController {
...
+  duration const PeriodicChecksInterval{50ms};
   duration const FileSizeCalcInterval{5s};
   time_point LastFileSizeCalcTime{system_clock::now() - FileSizeCalcInterval};
 
+  /// \brief Hysteresis factor to start refilling the write queue after a pause.
+  ///
+  /// Consumers are stopped when the write queue is larger than
+  /// StreamerOptions.MaxQueuedWrites. This variable defines the ratio below
+  /// which the consumers will be resumed.
+  float const QueuedWritesResumeThreshold{0.8F};
```
mattclarke, line 126: Good idea!

src/StreamController.h
```diff
@@ -61,6 +64,16 @@ class StreamController : public IStreamController {
   /// last message to be written in nanoseconds.
   void setStopTime(const time_point &StopTime) override;
 
+  /// \brief Pause consumers.
+  ///
+  /// Pauses consumer polling to throttle the ingestion of data.
+  void pauseStreamers() override final;
```
mattclarke, line 70: I'd probably not bother marking these methods as final. Very few other methods seem to have been marked final, so marking these two as final could be read as meaning there is something special about these two methods which I don't think is the case.

mattclarke (approved):
Looks good. 
I think the names changes and names for new items are a big improvement 👍

PR #698, opened 2023-07-03
ECDC-3494 honour hdf prefix setting

danesss:
## Issue

https://jira.esss.lu.se/browse/ECDC-3494

## Description of work

- Change the default storage directory for data files to be the current working directory, and not the root of the filesystem's mount point.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,9 +12,9 @@ To summarise:
 
 ### Branching
 
-- Branch your feature off 'master'
+- Branch your feature off 'main'
 
-- Create pull requests against 'master'.
+- Create pull requests against 'main'.
 
 ### Branch naming
 The names should start with the ticket number and contain a brief description. For example:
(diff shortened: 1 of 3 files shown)


ggoneiESS:
Line 83 in src/tests/MasterTests.cpp needs a check that filename isn't blank, otherwise you're going to end up deleting the folder and causing issues

& the comment on the CLI option is a bit confusing now - isn't it actually the relative path you're giving, not absolute?  That would make me use something like
hdf-output-prefix = "/ess/data/ymir/2022"

which means we would write to /whatever/folder/we/were/in/ess/data/ymir/2022/

so the comment should say 'addLineBreaks("Relative path to directory which gets "', no?

ggoneiESS:
> 

Do we need to check an absolute path has been given then?  It looks like it should be caught in FileWriterTask.cpp:58, but not 100% sure.  Might be cleaner to have a check, but I can't see any on any of the inputs

ggoneiESS:
> > isn't it actually the relative path you're giving, not absolute?
> 
> If the config option is unset, then we use current working directory. If the option is set, it should be set with an absolute path.

Do we need to check an absolute path has been given then? It looks like it should be caught in FileWriterTask.cpp:58, but not 100% sure. Might be cleaner to have a check, but I can't see any on any of the inputs

ggoneiESS (approved):
Looks good to me!

PR #703, opened 2023-08-14
Ecdc 3447 filewriter attribute module

ggoneiESS:
## Issue
ECDC-3447

## Description of work
Created a new module with four-byte identifier 'mdat' that currently processes start_time and stop_time and is expandable for future processing and other messages sent directly to Kafka.

The start time and stop time that already exist within the filewriter in other locations are written specifically at the correct times (in JobCreator.cpp where the start time is 'obvious' and StreamController.cpp, which is the method always processed regardless of filewriter state, except in the most catastrophic crashes).

Relevant tests have been added.

## Checklist

- [✓](https://fsymbols.com/signs/tick/) ] Changes have been documented in `changes.md`
- [ N/A ] Documentation has been updated in `documentation` directory


ggoneiESS:
I have no idea why this passed this time when I tried to rebase to an earlier commit, VScode complained, and I merged manually...dd337d7 and ec7eff8 should otherwise be identical!

conanfile.txt
ggoneiESS, line 1: These changes were needed to get a working env running locally - could (should?) be moved to different branch

src/StreamController.cpp
ggoneiESS, line 1: Start message is easy and can be done in command handling (to write to the 'head' of the NeXus file rather than a separate entry/block), but stop commands can come from the start message, other stop messages, or the program itself.  This is where all 'stops' eventually pass through however and is the right place to write to the file.

ggoneiESS, line 1: Could StreamController have an instance of mdat that adds a start and stop time (and/or other metadata) itself?

ggoneiESS: no - but we use a special method for the mdat module to make the writing simple

ggoneiESS:
This has the discussed changes in, with a new templated method writemetadata added to the mdat module.  This means the flatbuffer creation is removed, and instead the relevant info is passed directly to the method.  To achieve the call to the method, a static_cast is deployed but iff it is the mdat module so there is no need to worry about catching going into an undefined state.

mattclarke (approved):
Once Jenkins approves, feel free to merge

PR #706, opened 2023-08-31
ECDC-3394 Deflake integration tests

danesss:
## Issue

ECDC-3394

## Checklist

- [X] Changes have been documented in `changes.md`
- [ ] Documentation has been updated in `documentation` directory

### DIFF
--- a/integration-tests/test_large_start_message.py
+++ b/integration-tests/test_large_start_message.py
@@ -1,3 +1,5 @@
+import time
+
 from helpers.nexushelpers import OpenNexusFile
 from datetime import datetime, timedelta
 import pytest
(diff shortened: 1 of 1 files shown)


ggoneiESS (approved):
All code looks good, there are few changes that should have an effect on the actual functionality.

We should consider whether start times, stop times, and timeouts should be set as constant(s) elsewhere (e.g. SHORT_TIMEOUT, MEDIUM_TIMEOUT, LONG_TIMEOUT), and set in a more formal way based on the processing required from Filewriter, but this is out of scope of this which is needed for any branches to pass Jenkins consistently!

PR #711, opened 2023-09-28
stop_time to end_time and NX_DATE_TIME

ggoneiESS:
## Issue

Changed stop_time to end_time and changed the time format from nanoseconds to ISO8601 in line wiht NX_DATE_TIME

## Description of work
New method to account for the fact we are using a string rather than a time in the dataset (which is better in general anyway) and then use of <ctime> to do the conversion for formats; chrono fortmat options would be best but that means C++20

## Checklist

- [ x ] Changes have been documented in `changes.md`
- [ ] Documentation has been updated in `documentation` directory


ggoneiESS:
Let's discuss this tomorrow before the quick catch-up and the code hour?  0930?  Writing NX_DATE_TIME as strings is actually quite a different puzzle compared to before, since it uses a very different NeXus data type.  I also don't understand how we will write times (with timezones or UTC?)

src/HDFOperations.cpp
```diff
@@ -246,6 +246,9 @@ void writeStringDataset(hdf5::node::Group const &Parent,
     auto DataType = hdf5::datatype::String::variable();
     DataType.encoding(hdf5::datatype::CharacterEncoding::UTF8);
     DataType.padding(hdf5::datatype::StringPad::NullTerm);
+    std::cout << "JsonHERE" << std::endl
+              << Values.dump() << std::endl
+              << std::endl;
```
mattclarke, line 251: delete this.

src/JobCreator.cpp
```diff
@@ -187,10 +187,8 @@ createFileWritingJob(Command::StartInfo const &StartInfo, MainOpt &Settings,
   for (auto &Item : Task->sources())
     if (Item.writerModuleID() == "mdat")
       static_cast<WriterModule::mdat::mdat_Writer *>(Item.getWriterPtr())
-          ->writemetadata("start_time",
-                          std::chrono::duration_cast<std::chrono::milliseconds>(
-                              StartInfo.StartTime.time_since_epoch())
-                              .count());
+          ->writemetadata("start_time", StartInfo.StartTime);
+  //                          std::chrono::duration_cast<std::chrono::milliseconds>(StartInfo.StartTime.time_since_epoch()).count());
```
mattclarke, line 191: delete this.

